PR #1, opened 2019-12-02
Benchmarking workflow initial check-in

kachulis:
@yfarjoun could you approve initial check-in.  Once it's in we can start applying updates here.

### DIFF
--- a/BenchmarkVCFs/BenchmarkVCFs.wdl
+++ b/BenchmarkVCFs/BenchmarkVCFs.wdl
@@ -48,6 +48,8 @@ workflow Benchmark {
         stratLabels: {description: "labels by which to identify stratification intervals (must be same length as stratIntervals)"}
         jexlVariantSelectors: {description: "variant types to select over (defined by jexl fed to GATK SelectVariants)"}
         variantSelectorLabels: {description: "labels by wich to identify variant selectors (must be same length as jexlVariantSelectors)"}
+        doIndelLengthStratification: {description: "whether or not to perform stratification by indel length"}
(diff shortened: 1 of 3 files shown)


yfarjoun:
should this have a test json? what do you think our standards should be? I was thinking that we should and have womtool validate with input using circle-ci....if you put in a json, I'll setup the CI part.

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -0,0 +1,1336 @@
+version 1.0
+
```
yfarjoun: since this is public, let's stick a blurb about what this does.

yfarjoun:
let's have a readme.md and an (MIT) license

kachulis:
I put a bsd-3 license on the entire repo (it’s what comms uses for the
gatk-workflows repos).  Do you have a preference for one license over the
other?  And do you think it’s necessary to have individual licenses on each
workflow?

yfarjoun:
bsd-3 seems fine to me. no need for license on individual files in that
case.

BenchmarkVCFs/BenchmarkVCFs.wdl
kachulis: added as part of a meta section

kachulis:
I think some form of automated testing is definitely a good idea.  `Womtool validate` at the minimum, IMO; In the long run it would be nice to have more sophisticated testing with plumbing data actually running through the workflows.  I have added a test.json if we want to just start with the easy `womtool validate` option.

PR #2, opened 2019-12-03
Added an option for "analysis region" which is for when the whole ana…

yfarjoun:
…lysis is happening over a subset (say chr9 or chr20:10000000-20000000) but you dont' want to have to prepare all the stratification ILs yourself.

### DIFF
--- a/BenchmarkVCFs/BenchmarkVCFs.wdl
+++ b/BenchmarkVCFs/BenchmarkVCFs.wdl
@@ -21,11 +21,13 @@ workflow Benchmark {
         Array[String]? variantSelectorLabels
         String referenceVersion
         Int? threadsVcfEval=2
-        Int? threadsHappy=4
         Boolean doIndelLengthStratification=true
         Int? preemptible
         String gatkTag="4.0.11.0"
         Boolean requireMatchingGenotypes=true
+
+        Boolean passingOnly=true
+        String? vcfScoreField
     }
 
     meta {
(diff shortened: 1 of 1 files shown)


BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -20,15 +21,14 @@ workflow Benchmark {
         Array[String]? variantSelectorLabels
         String referenceVersion
         Int? threadsVcfEval=2
+        Int? threadsHappy=4
```
kachulis: Workflow currently doesn't run happy or GenotypeConcordance, only vcfeval.  If we want to include happy we should add the call to the task before adding this parameter.  Personally I would prefer to avoid adding happy, as it's much slower and doesn't seem to be any better than vcfeval.

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -50,6 +50,22 @@ workflow Benchmark {
         variantSelectorLabels: {description: "labels by wich to identify variant selectors (must be same length as jexlVariantSelectors)"}
         doIndelLengthStratification: {description: "whether or not to perform stratification by indel length"}
         requireMatchingGenotypes: {description: "whether to require genotypes to match in order to be a true positive"}
+        threadsHappy: {description: "threads (and cpu cores) to use for hap.py. Defaults to 4."}
```
kachulis: No happy

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -50,6 +50,22 @@ workflow Benchmark {
...
+        gatkTag: {description: "version of gatk docker to use.  Defaults to 4.0.11.0"}
+        happyTag: {description: "version of hap.py docker to use.  Defaults to v0.3.9"}
```
kachulis: No happy

kachulis:
@yfarjoun it looks like this PR just expanded a bit (namely the addition of roc curve stuff).  Is that your intention?

yfarjoun:
yes...though you may call this scope creep...

PR #3, opened 2019-12-03
- first attempt at testing

yfarjoun:


.circleci/config.yml
```diff
@@ -0,0 +1,22 @@
+# CircleCI 2.0 configuration file
+#
+#
+version: 1
```
kachulis: can we use circleci 2.0 instead?

.circleci/config.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  build:
+    docker:
+      # specify the version you desire here
+      - image: broadinstitute/womtool:47-b36d920
+        entrypoint: /bin/bash
+    branchs:
```
kachulis: branches

yfarjoun: (I guess it didn't do anything...)

.circleci/config.yml
yfarjoun: 👍

PR #5, opened 2019-12-06
Testing by running with plumbing data

kachulis:


BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -1397,6 +1401,21 @@ task MatchEvalTruth {
...
 
+    parameter_meta {
+    	evalVcf: {
+    		localization_optional: true
+    	}
+    	truthVcf: {
+    		localization_optional: true
+    	}
+    	evalVcfIndex: {
+			localization_optional: true
```
yfarjoun: tabs are funny

.circleci/config.yml
```diff
@@ -1,20 +1,79 @@
...
         entrypoint: /bin/bash
     working_directory: ~/palantir-workflows
 
-    environment:
-      # Customize the JVM maximum heap limit
+commands:
+  get-cromwell-jar:
+    description: "Download cromwell jar"
+    steps:
+      - run: wget https://github.com/broadinstitute/cromwell/releases/download/47/cromwell-47.jar
```
yfarjoun: you could also run the cromwell jobs in a different "job" that has the other docker (the cromwell one)

kachulis:
@yfarjoun moved back to Makefile based structure, so back to you!

Makefile
```diff
@@ -1,22 +1,28 @@
...
+validate: $(all-validations)
 
-%.test : %.wdl
-	$(VALIDATE)  $?
+.PHONY: %.wdl.validate
+%.wdl.validate : 
+	$(VALIDATE) $(basename $@)
 
-%.test.json : %.wdl.json
-	$(VALIDATE) $(basename $? .json) -i $?
```
yfarjoun: I think it's good to have a validation step that is a "full" validation, i.e. that the json supplies all the required values, ifcourse this will be de-facto tested by the test target, but then the error will be less clear. is it difficult to have three levels of testing? (validation w/o json, validation w/ json, testing with json and comparison to expected results?

.circleci/config.yml
kachulis: Since running cromwell results in calling docker, circleci requires treating differently if done in a docker (ie, https://circleci.com/docs/2.0/building-docker-images/).  Among other things, this requires installing docker manually.  All in all, doesn't seem worth it to me.

Makefile
kachulis: Good point, added validation with jsons before running full tests

kachulis:
@yfarjoun back to you.  Shouldn't actually be that much changed, most of those commits are messing around with docker-in-docker madness for the `test` job, which I eventually reverted back to the way I had it originally.

yfarjoun:
This is good. one complaint remaining...the test output is so verbose, I have no idea what's going on....can you redirect the logs of each job to its own artifact (and publish those artifacts)?

Makefile
```diff
@@ -31,7 +31,7 @@ validate-with-json: $(all-validations-with-json)
 .PHONY: %.json.test
 %.json.test:
 	mkdir -p logs
```
yfarjoun: set -o pipefail?

kachulis: circleci runs by default with `-eo pipefail`

PR #7, opened 2020-05-22
first draft imputation pipeline

skwalker:


ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+task QConChunk {
+  input {
+    File vcf
+    File vcf_index
+    File panel_bcf
+    File panel_bcf_index
+    File sarah_jar
+    Int disk_size = 400
+  }
+  command <<<
```
skwalker, line 240: this is disgusting and needs to be improved :)

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    File input_vcf
+    File input_vcf_index
+    String output_vcf_basename
+   }
+    Int disk_size = 400
+
+  command <<<
+    # site missing rate < 5% ; hwe p > 1e-6 ; in
+    vcftools --gzvcf ~{input_vcf} --stdout --max-missing 0.05 --hwe 0.000001 --recode -c > tmp.vcf
+    awk 'BEGIN {FS="\t"}; {if ($1 ~ /#/ || ($4=="A" && $5!="T") || ($4=="T" && $5!="A") || ($4=="C" && $5!="G" ) || ($4=="G" || $5!="C")) print $0}' tmp.vcf | bgzip -c > ~{output_vcf_basename}.vcf.gz
```
skwalker, line 462: probably want to fix this  too

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+task MergeVCFs {
+  input {
+    Array[File] input_vcfs
+    String output_vcf_basename
+    Float shard_size
+  }
+  Int disk_size = ceil(((shard_size * 22) * 2) + 20)
+  command <<<
+    valid_vcfs=$(for file in ~{sep=' ' input_vcfs}; do rows=$(zgrep -v "#" $file | wc -l); if [ $rows -gt 0 ]; then echo $file; fi; done)
+    bcftools concat $valid_vcfs | bgzip -c > ~{output_vcf_basename}.vcf.gz
```
kachulis, line 350: -O instead of piping to bgzip.

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  }
+}
+
+task MergeVCFs {
+  input {
+    Array[File] input_vcfs
+    String output_vcf_basename
+    Float shard_size
+  }
+  Int disk_size = ceil(((shard_size * 22) * 2) + 20)
```
kachulis, line 347: compute based on summed size of input array, not estimated scaling of shard_size

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+      cpu: "1"
+        disks: "local-disk 100 HDD"
+  }
+}
+
+task MergeVCFs {
+  input {
+    Array[File] input_vcfs
+    String output_vcf_basename
+    Float shard_size
```
kachulis, line 345: don't need shard_size

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  }
+  Int disk_size = ceil(((shard_size * 22) * 2) + 20)
+  command <<<
+    valid_vcfs=$(for file in ~{sep=' ' input_vcfs}; do rows=$(zgrep -v "#" $file | wc -l); if [ $rows -gt 0 ]; then echo $file; fi; done)
```
kachulis, line 349: I think you need to do this because a chunk which had no sites after imputation wouldn't have any samples listed in it's header, which would then mean that bcftools will compain that is has different samples than all the other files (which is annoying).

This could be sped up by instead just searching for the line which specifies the column names, and checking that there are more than 9 fields.  This way you don't need to feed the entire file into grep and wc.

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  Array[File] phased_vcfs = flatten(minimac4.vcf)
+
+  call MergeVCFs {
+    input:
+      input_vcfs = phased_vcfs,
+      output_vcf_basename = output_callset_name,
+      shard_size = size(minimac4.vcf[0], "GB")
+  }
+
+  File michiganserver_vcf = MergeVCFs.output_vcf
```
kachulis, line 104: this isn't Michigan!

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+}
+
+task UpdateHeader {
+  input {
+    File vcf
+    File ref_dict
+    Int disk_size = 400
+  }
+  command <<<
+  zgrep -v "##contig=<ID=" ~{vcf} | bgzip -c > no_contigs.vcf.gz
```
kachulis, line 369: why is this needed?  Isn't the `UpdateVCFSequenceDictionary` line just going to remove and replace all these anyway?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  output {
+    File output_vcf = "~{output_vcf_basename}.vcf.gz"
+  }
+}
+
+task UpdateHeader {
+  input {
+    File vcf
+    File ref_dict
+    Int disk_size = 400
```
kachulis, line 366: dynamic

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    File output_vcf_index = "final_call.vcf.gz.tbi"   
+  }
+}
+
+task RemoveSymbolicAlleles {
+  input {
+    File original_vcf
+    File original_vcf_index 
+    String output_basename = "michigan_allchroms.no_symbolic"
+    Int disk_size = 200
```
kachulis, line 390: dynamic

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    File output_vcf = "final_call.vcf.gz"
+    File output_vcf_index = "final_call.vcf.gz.tbi"   
+  }
+}
+
+task RemoveSymbolicAlleles {
+  input {
+    File original_vcf
+    File original_vcf_index 
+    String output_basename = "michigan_allchroms.no_symbolic"
```
kachulis, line 389: Michigan?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  }
+  command {
+    gatk SelectVariants -V ~{original_vcf} -xl-select-type SYMBOLIC --select-type-to-exclude MIXED --exclude-non-variants TRUE --remove-unused-alternates TRUE -O ~{output_basename}.vcf.gz
+  }
+  output {
+    File output_vcf = "~{output_basename}.vcf.gz"
+    File output_vcf_index = "~{output_basename}.vcf.gz.tbi"
+  }
+  runtime {
+   # docker: "us.gcr.io/broad-gatk/gatk:latest"
```
kachulis, line 400: remove

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+
+  File michiganserver_vcf = MergeVCFs.output_vcf
+
+  call UpdateHeader {
+    input:
+      vcf = michiganserver_vcf,
+      ref_dict = ref_dict
+  }
+
+  call RemoveSymbolicAlleles as RSAlleles {
```
kachulis, line 112: why the special alias?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    memory: "4 GB"
+  }
+}
+
+task SeparateMultiallelics {
+  input {
+    File original_vcf
+    File original_vcf_index 
+    String output_basename = "michigan_allchroms.no_multi_symbolic"
+    Int disk_size = 200
```
kachulis, line 412: dynamic

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    disks: "local-disk " + disk_size + " HDD"
+    memory: "4 GB"
+  }
+}
+
+task SeparateMultiallelics {
+  input {
+    File original_vcf
+    File original_vcf_index 
+    String output_basename = "michigan_allchroms.no_multi_symbolic"
```
kachulis, line 411: Michigan again!

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  command {
+    bcftools norm -m - ~{original_vcf} | bgzip -c > ~{output_basename}.vcf.gz
+    bcftools index -t ~{output_basename}.vcf.gz
+  }
+  output {
+    File output_vcf = "~{output_basename}.vcf.gz"
+    File output_vcf_index = "~{output_basename}.vcf.gz.tbi"
+  }
+  runtime {
+    docker: "skwalker/imputation:with_python"
```
kachulis, line 423: official docker?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    memory: "4 GB"
+  }
+}
+
+task SplitSample {
+  input {
+    File vcf
+    File vcf_index
+    String sample
+    Int disk_size = 400
```
kachulis, line 434: dynamic

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  }
+  command {
+    gatk SelectVariants -V ~{vcf} -sn ~{sample} -O ~{sample}.unsorted.vcf
+    gatk --java-options "-Xmx8g" SortVcf -I ~{sample}.unsorted.vcf -O ~{sample}.vcf.gz
```
kachulis, line 438: can these be piped together to save disk space, and probably also speed?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+  command <<<
+    bcftools norm -d both ~{input_vcfs} | bgzip -c > ~{output_basename}.vcf.gz
+    bcftools index -t ~{output_basename}.vcf.gz
+  >>>
+  runtime {
+    docker: "farjoun/impute:0.0.3-1504715575"
+    memory: "3 GiB"
+    disks: "local-disk " + disk_size + " HDD"
+  }
+  output {
```
kachulis, line 155: are there any metrics from minimac or eagle, or the qc of the input arrays that might be good to have as outputs of the workflow?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
+version 1.0
+
+workflow TestingNewImputation {
+  input {
+    Int chunkLength = 25000000
+    Array[File] array_vcfs
+    Array[File] array_vcf_indices 
+    Array[String] samples
+    Boolean perform_qc_steps
```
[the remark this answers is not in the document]
edytamalolepsza: this step was supposed to work well when imputing multiple samples

ImputationPipeline/Imputation.wdl
skwalker: It shouldn't be necessary to do both ( I think this was some issue I was having in like January with gzipped vs normal vcfs but I think you might have actually fixed this ?)

kachulis: ohhh right I remember, it was this bug: broadinstitute/gatk#6443

that is (for our purposes) fixed in gatk now, so SelectVariants can output a gzipped vcf, and the output will be correctly sorted so you don't need the call to `SortVcf`

ImputationPipeline/Imputation.wdl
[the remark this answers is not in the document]
skwalker: The QC stuff here is not the Michigan QC steps, but the steps from Amit, which are often based on sample specific metrics so if you were to impute one or only a couple of samples you wouldn't want to use them

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    call CalculateChromsomeLength {
+      input:
+        ref_dict = ref_dict,
+        chrom = chrom
+    }
+    
+    Float chunkLengthFloat = chunkLength
+    Int num_chunks = ceil(CalculateChromsomeLength.chrom_length / chunkLengthFloat)
+
+    scatter (i in range(num_chunks)) {
```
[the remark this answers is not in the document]
skwalker: Interesting, I've never used that tool before. One thing I'm a little concerned about it is that to be most similar to Michigan, we would want the chunks to be exactly 25,000,000 bp long like Michigan, but I don't think you can specify the interval split length, only the number of intervals (unless I'm reading the documentation incorrectly)

ImputationPipeline/Imputation.wdl
skwalker: I think Edyta was using the job twice, this is unnecessary now, thanks for catching 👍

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+
+workflow TestingNewImputation {
+  input {
+    Int chunkLength = 25000000
+    Array[File] array_vcfs
+    Array[File] array_vcf_indices 
+    Array[String] samples
+    Boolean perform_qc_steps
+    File ref_dict = "gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.dict"
+    File new_gatk_jar = "gs://broad-dsde-methods-skwalker/polygenic_risk_scores/test_new_gatk.jar"
```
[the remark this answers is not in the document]
skwalker: ugh I can't remember anymore why I did that :( I guess we can see what happens using normal GATK

ImputationPipeline/Imputation.wdl
skwalker: I think Michigan just outputs which variants passed their QC thresholds, so I guess we could add that
There's the "info" file from Minimac which we output, although don't save as a final output of the workflow, it tells you some info about each site being imputed I think? I'm not sure how helpful it is

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+    File reference_panel_bcf_index
+    String chrom
+    Boolean valid_chunk
+    String genetic_map_file
+    Int start
+    Int end
+  }  
+  Int disk_size = 400
+  command <<<
+    if [ ~{valid_chunk} = "true" ]; then
```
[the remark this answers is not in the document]
skwalker: Yeah, I think I had an issue in the final gather after the scatter but I will try again

ImputationPipeline/Imputation.wdl
skwalker: I think Yossi's docker was using an older version of bcftools that didn't have that option but I will update to a better bcftools docker

ImputationPipeline/Imputation.wdl
skwalker: Michigan does this in their QC because they aren't be used in the phasing/imputation algorithms and I wasn't sure if it would break them to give them the data or if they would ignore it properly?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,475 @@
...
+             --bpStart ~{start} --bpEnd ~{end} --allowRefAltSwap 
+    else 
+        bcftools view ~{dataset_bcf} -Oz -o pre_phased_~{chrom}.vcf.gz
+    fi
+  >>>
+  output {
+    File dataset_prephased_vcf="pre_phased_~{chrom}.vcf.gz"
+  }
+  runtime {
+    docker: "skwalker/imputation:test"
```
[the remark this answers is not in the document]
skwalker: I have the docker file somewhere for this... We wanted it to match exactly the version being used by the Michigan server

ImputationPipeline/Imputation.wdl
edytamalolepsza: this reminds me that we need to decide what files to output

ImputationPipeline/Imputation.wdl
skwalker: Wait I'm sorry.. I misread their documentation, but looking at the code they do allow for ambiguous SNPs

https://github.com/genepi/imputationserver/blob/57445da98cfeff9cb69978b08547d885d1bcccbb/src/main/java/genepi/imputationserver/steps/fastqc/StatisticsTask.java

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,453 @@
...
+
+  call GenerateDataset {
+    input:
+      input_vcfs = array_vcfs,
+      input_vcf_indices = array_vcf_indices,
+      output_vcf_basename = "merged_aou_arrays"
+  }
+
+  if (perform_qc_steps) {
+        call QConArray {
```
skwalker, line 25: I'm thinking this task should be combined with QC on Chunk or at least called on the chunk level instead (it'll be faster / less expensive this way ) -- @kachulis , @edytamalolepsza: thoughts?

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,453 @@
...
+  input {
+    File vcf
+    File vcf_index
+    File panel_vcf
+    File panel_vcf_index
+    Int disk_size = size([vcf, vcf_index, panel_vcf, panel_vcf_index], "GB")
+  }
+  command <<<
+
+    ### IMPORTANT TODO: make sure ref and alts match between array and panel or are a simple mismatch
```
skwalker, line 237: I think we can maybe save this for later because it's going to take a while

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,453 @@
...
+            prefix = "chrom" + "_chunk_" + i +"_imputed",
+            chrom = chrom,
+            start = (i * chunkLength) + 1,
+            end = (i + 1) * chunkLength
+        }
+      }
+     }
+  }
+
+  Array[File] phased_vcfs = flatten(select_all(minimac4.vcf))
```
skwalker, line 92: @kachulis : how would I get this to work? I'm getting this error now:

`Failed to process workflow definition 'TestingNewImputation' (reason 1 of 1): Failed to process declaration 'Array[File] phased_vcfs = flatten(select_all(minimac4.vcf))' (reason 1 of 1): Failed to process expression 'flatten(select_all(minimac4.vcf))' (reason 1 of 1): Invalid parameter 'IdentifierMemberAccess(minimac4,vcf,Vector())'. Expected an array of optional values (eg 'Array[X?]') but got 'Array[Array[File?]]'`

skwalker: oh actually I got it to work now ( I think )

ImputationPipeline/Imputation.wdl
```diff
@@ -0,0 +1,453 @@
...
+    File panel_vcf
+    File panel_vcf_index
+    Int disk_size = size([vcf, vcf_index, panel_vcf, panel_vcf_index], "GB")
+  }
+  command <<<
+
+    ### IMPORTANT TODO: make sure ref and alts match between array and panel or are a simple mismatch
+ 
+    var_in_original=$(gatk CountVariants -V ~{vcf})
+    var_in_original=$(gatk CountVariants -V ~{vcf} -L ~{panel_vcf})
```
kachulis, line 240: I think this should be `var_in_reference`

edytamalolepsza: It is, Sarah corrected it already

kachulis: hmm, when I look at the latest commit it is still `var_in_original`... unless I am missing something.

I mean here in the git repo, not on a Terra snapshot.

ImputationPipeline/Imputation.wdl
kachulis: I would lean toward keeping QConChunk and QConArray separate, since they are really doing pretty different thing (one removing individual sites, the other deciding if a particular chunk of genome can be phased/imputed).  But I do agree that moving this into the scattered part of the pipeline could be a nice speedup.  I actually don't think it will make things less expensive though, since you still need to do the same amount of computation, just scattered across a number of machines in parallel.  It might even be slightly more expensive since there will be duplication of overhead, but probably worth it if this is a speed bottleneck.

ImputationPipeline/Imputation.wdl
edytamalolepsza: I see... (Terra version is corrected)

PR #12, opened 2020-09-29
allow-extra-chr when running plink

kachulis:
@edytamalolepsza please approve once you confirm that this works and we can cut a new release.

### DIFF
--- a/ImputationPipeline/PerformPopulationPCA.wdl
+++ b/ImputationPipeline/PerformPopulationPCA.wdl
@@ -120,6 +120,7 @@ task LDPruning {
     --extract ~{original_array_sites} \
     --indep-pairwise 1000 50 0.2 \
     --maf 0.01 \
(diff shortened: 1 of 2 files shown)


edytamalolepsza:
It passed the extra chromosomes, but failed with the following message:
```
--vcf: 291241k variants converted.Warning: 6630150 --score file entries were skipped due to missing variant IDs.
Error: No valid variants in --score file.
```

edytamalolepsza:
I think the lift over to hg37 did not generate correct file, please see the `chr` still present in the ID column:

`1       10481   chr1:10481:A:G  A       G       .       PASS    AC=0;AF=0.00;AN=2;IMPUTED;MAF=0.00000;R2=0.00001        GT:DS:GP:HDS    0|0:0:1,0,0:0,0`

kachulis:
superseded by #13

PR #23, opened 2021-01-16
making the BenchmarkVcf shiny app public

yfarjoun:
I moved this from our private "palantir" repository and made a few changes so that it works in the presence of multiple inputs (vcfs) and multiple interval lists. 

would be nice to (eventually) publish this directly from github using a more generic credential, rather than it being on my personal account.....


meganshand (approved):
I couldn't get this to run locally (it freezes when I try to load a file?), but I could deploy it on shinyapps.io/shinyapp and it works there, which I think is good enough for now.

PR #8, opened 2020-07-22
Dockstore's Github App has arrived!

kachulis:
^^

PR #24, opened 2021-01-21
Update dockstore yml to include Benchmarking script

meganshand:
I now see this branch in the BenchmarkVCFs dockstore page (but none of the other branches). Perhaps once this is merged to main it will update all the branches? I'm not sure how to test that theory without merging it, but if someone else has ideas please let me know.


kachulis:
@meganshand each branch is handled based on the yml on that branch, so branches which include the correct yml lines will automatically sync, and those which don't will not.  

this looks good, but closing since you discovered the forgotten #8, lol

PR #25, opened 2021-01-26
Use gcr instead of dockerhub in Benchmarking wdl

meganshand:
I was getting timeout errors from pulling the gatk docker from dockerhub. Using google instead should fix that.

### DIFF
--- a/BenchmarkVCFs/BenchmarkVCFs.wdl
+++ b/BenchmarkVCFs/BenchmarkVCFs.wdl
@@ -479,7 +479,7 @@ task CheckForVariants {
     >>>
 
     runtime {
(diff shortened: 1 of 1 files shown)


kachulis:
👍  yeah the dockerhub throttling strikes again.  Should we move all the dockers over to gcr?  I imagine this will affect anything that scattered any significant number of times...

meganshand:
eh, I wouldn't worry until it becomes an issue. It's always easy to switch over later, it's just especially nice that gatk already exists on gcr.

meganshand:
Oh @kachulis could you please officially approve? Github won't let me merge without an "approving review"

kachulis:
sounds good!

kachulis (approved):

PR #29, opened 2021-03-26
Tm standard docker images

tmelman:
This PR replaces Docker images in the following ways:
* Replaces Dockerhub hosted images with GCR hosted images (to avoid extra charges)
* Replaces personal Docker images with ones for which we have the Dockerfile 

I also added scripts for building and pushing the dockerfiles, at least so we have documentation about how they were built and where they were pushed to. 

This change was tested in the following way:
* Ran the standard pipeline and this updated pipeline on Terra using the same inputs
* Picked a subset of the VCF data (chr20) from each output
* `diff`ed the VCFs, made sure there are no diffs

### DIFF
--- a/.dockstore.yml
+++ b/.dockstore.yml
@@ -15,3 +15,12 @@ workflows:
   - name: BenchmarkVCFs
     subclass: WDL
     primaryDescriptorPath: /BenchmarkVCFs/BenchmarkVCFs.wdl
+  - name: ValidateImputation
+    subclass: WDL
+    primaryDescriptorPath: /ImputationPipeline/Validation/ValidateImputation.wdl
+  - name: ValidateScoring
+    subclass: WDL
+    primaryDescriptorPath: /ImputationPipeline/Validation/ValidateScoring.wdl
+  - name: FullImputationPRSValidation
+    subclass: WDL
+    primaryDescriptorPath: /ImputationPipeline/Validation/FullImputationPRSValidation.wdl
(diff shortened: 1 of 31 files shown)


kachulis:
@tmelman this looks great!  Could you just merge in changes from main, and rerun the check for differences?  I want to make sure the output qc metrics changes get merged into this correctly.

kachulis:
also since the `build_and_push_docker.sh` script can be used for any dockers in the repo, let's move it it's own top level directory (DockerUtils or something) so that it's not associated purely with the imputation workflow.

kachulis:
@tmelman now that the validation workflows are in main can you merge changes in main into this branch and run the validation workflows for this branch, and post results here?

tmelman:
image

kachulis:
@tmelman do the Pearson Correlation plots also show no change?  I'd also like to see that the metrics output from the imputation pipeline remain the same.  And I wonder if we can md5 the imputed vcfs and compare, or if the headers will have changed?

tmelman:
Here are the pearson correlations:

image

tmelman:
I `diff`ed the output VCF files (selecting chromsome 20 for efficiency) and found no differences.

tmelman:
I diffed the two metrics files (`broad_imputation_aggregated_imputation_metrics.tsv`) and found no differences.

kachulis (approved):
Looks good @tmelman!  And thanks for all of the additional checks!

PR #31, opened 2021-03-29
Validation Workflows for Imputation/PRS

kachulis:


kachulis:
Adds two validation workflows, one to validate scoring, the other to validate imputation, and a third workflow to stitch them together.  These produce three main plots.  The first is and r^2 vs af calculation, to evaluate imputation quality:
image
if something has actually changed between the branch and main, the two lines might be distinguishable.  in this case the only difference from main for this branch is the addition of the two workflows, so no change in performance

Next is a comparison between imputed prs and wgs prs:
image

Finally is a comparison of imputed prs on this branch and on the main branch:
image

ImputationPipeline/Validation/ValidateImputation.wdl
```diff
@@ -0,0 +1,315 @@
+version 1.0
```
tmelman, line 1: ValidateImputation looks good to me

ImputationPipeline/Validation/ValidateScoring.wdl
```diff
@@ -0,0 +1,264 @@
...
+		input:
+			sites = extractImputedIDsMain.ids,
+			weights = weights
+	}
+
+
+	call Scoring.ScoringImputedDataset as ScoreImputed {
+		input:
+			weights = weights,
+			imputed_array_vcf = select_first([validationArrays, validationArraysMain]),
```
tmelman, line 58: Why would validationArrays be optional (and fall back to the one in main) but validationArraysMain not?

ImputationPipeline/Validation/ValidateScoring.wdl
```diff
@@ -0,0 +1,264 @@
...
+			redoPCA = true
+	}
+
+	call QCSites {
+		input:
+			input_vcf = validationWgs,
+			output_vcf_basename = "wgsValidation"
+	}
+
+	call Scoring.ScoringImputedDataset as ScoreWGS {
```
tmelman, line 89: It would be helpful to have some comments throughout both these WDLs; here, for example, to make clear what the difference between ScoringImputedMain and ScoringWGS is (since it's not just 2 branches).

Highest priority would be comments explaining what the inputs are.

tmelman (approved):
My comments are low-priority, I approve merging.

ImputationPipeline/Validation/ValidateScoring.wdl
kachulis: we need at least on set input array vcf to test on.  There are two basic testing setups I'm trying to cover here:
1) We want to test just the difference in the scoring pipeline between the new branch and main.  In this case, we want to run the scoring on the same vcf with both branches.  It doesn't really matter whether the shared input in this case is validationArrays or validationArraysMain, I just picked one.
2) We want to test the differences that changes to both the imputation and scoring pipelines together (or more likely, changes to imputation with unchanged scoring) have on prs scores.  In this case we need to have two different inputs for the results of imputation on the two different branches.

ImputationPipeline/Validation/ValidateScoring.wdl
kachulis: good call, I will add some comments

ImputationPipeline/ScoringPart.wdl
```diff
@@ -27,14 +28,20 @@ workflow ScoringImputedDataset {
...
   
   call ScoreVcf as ScoreImputedArray {
   	input:
   	vcf = imputed_array_vcf,
   	basename = basename,
   	weights = weights,
   	base_mem = scoring_mem,
-  	extra_args = columns_for_scoring
+  	extra_args = columns_for_scoring,
+  	sites = ExtractIDsPopulation.ids
```
kachulis, line 44: This was needed from #30.  Otherwise if there are sites in the "imputed" array which are not in the population array the scoring adjustment will not work correctly.

PR #33, opened 2021-04-06
split multisample vcf into single sample vcfs at end of imputation pipeline

kachulis:


kachulis:
Also adds crosscheck of output imputed data against input data to confirm no sample swaps have occurred in the pipeline.

Tested with splitting on 42 sample cohort, and without splitting (so testing only the addition of the crosscheck) on 13,000 sample cohort.  Crosscheck scales completely fine, negligible addition to computation costs of the pipelin.  Not sure how splitting will scale, particularly how Terra UI will respond to a massive array of files as a pipeline output.  Something we will need to test before running with that option turned on for large cohorts, but I don't think we should wait on that to merge this PR.

tmelman (approved):
LGTM

PR #35, opened 2021-05-13
Replaced adjustscores call to google bucket script with Heredoc

tmelman:
Moved R script contents to a "heredoc". 

Tested functional equivalent by diffing output from scoring pipeline in main vs in this branch.

### DIFF
--- a/ImputationPipeline/ScoringPart.wdl
+++ b/ImputationPipeline/ScoringPart.wdl
@@ -245,7 +245,6 @@ task ProjectArray {
 # This does the scoring adjustment
 task AdjustScores {
 	input {
-#		File adjusting_Rscript
 		File population_pcs
 		File population_scores 
 		File array_pcs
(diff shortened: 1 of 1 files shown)


ImputationPipeline/ScoringPart.wdl
```diff
@@ -254,10 +254,63 @@ task AdjustScores {
...
+	# Rscript ~{adjusting_Rscript} ~{population_pcs} ~{population_scores} ~{array_pcs} ~{array_scores}
+	command <<<
+		Rscript <<- "EOF"
+			library(ggplot2)
+			library(dplyr)
 
-	command {
-		Rscript ~{adjusting_Rscript} ~{population_pcs} ~{population_scores} ~{array_pcs} ~{array_scores}
-	}
+			args = commandArgs(trailingOnly=TRUE)
```
kachulis, line 263: I don't think this will work, since there aren't any args being passed.  But I think we should remove the use of command line args here, since we can just reference the task inputs directly in the R code, instead of passing them in through the command line.

tmelman:
The branch has been tested against `main` via the validation pipeline and shows functional equivalence.

image

kachulis (changes requested):

tmelman:
image

kachulis (approved):
looks good!

PR #38, opened 2021-05-19
Fixes bug in validation workflow

kachulis:
@tmelman this was the bug that used a "<" instead of "<=" in the if statement


tmelman (approved):
LGTM

PR #47, opened 2021-07-06
BuildHaplotypeMap

mollysacks:
Workflow that takes in a multisample & genotyped VCF, a header file, and various pruning parameters and builds a haplotype map consisting of single sites that are in low LD with each other


HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
+# BuildHaplotypeMap
+#
+# This workflow creates a haplotype map file containing a set of low LD sites for fingerprinting
+#
+# Inputs:
+#
+# 1. Genotyped multisample GVCF
```
kachulis, line 7: does it need to be a GVCF, or will a VCF work?

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+
+	command <<<
+python <<CODE
+import pandas as pd
+import vcf
+
+
+vcf_file = "${recode_vcf}"
+prune_in = "${prune_in}"
+header = open("${header_file}")
```
kachulis, line 74: use `with` statements so you don't need to close these yourself (and get guaranteed closing without using a try-finally block)

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+    File input_vcf
+    
+    command <<<
+  	vcftools --vcf ${input_vcf} --max-missing-count 0 --remove-indels --recode --recode-INFO-all --out ${intermediates_prefix}
+    >>>
+	
+    runtime {
+    disks: "local-disk 1000 HDD"
+    memory: "3500 MB"
+    docker: "jeltje/vcftools"
```
kachulis, line 26: use biocontainers/vcftools instead (it's more standard/maintained)

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+	output {
+    File recode_vcf = "${intermediates_prefix}.recode.vcf"
+  }
+}
+
+task plink {
+  File input_vcf
+  String intermediates_prefix
+  String output_prefix
+  Int prune_window
```
kachulis, line 38: these pruning related values should be optional with some reasonable default values

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+  String set_missing = "@:#"
+
+  command <<<
+  	plink --vcf ${input_vcf} --snps-only --maf ${min_maf} --biallelic-only --set-missing-var-ids ${set_missing} --indep-pairwise ${prune_window} ${prune_slide} ${prune_cutoff} --out ${intermediates_prefix}
+    >>>
+
+  runtime {
+    disks: "local-disk 1000 HDD"
+    memory: "3500 MB"
+    docker: "jrose77/plinkdocker"
```
kachulis, line 52: biocontainers/plink1.9

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+import pandas as pd
+import vcf
+
+
+vcf_file = "${recode_vcf}"
+prune_in = "${prune_in}"
+header = open("${header_file}")
+new_map = open("${output_prefix}.hapmap.txt", 'w')
+vcf_reader = vcf.Reader(open(vcf_file, 'r', encoding='utf-8'))
+vcf_chr = False
```
kachulis, line 77: hmmm, I don't like this whole vcf_chr thing... why is this necessary?  The snp ids should have been created based on the filtered vcf, which is what is also being used here, so why would this difference occur?

Regardless, I think an easier way to do this would be to annotate the vcf with IDs before passing it to plink (you can do `bcftools annotate --set-id '%CHROM\:%POS\:%REF\:%FIRST_ALT'`).   Then you don't have to worry about setting missing ids in plink, checking for 'chr', or reassigning IDs in this python script.

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+first = next(vcf_reader)
+if str(first.CHROM)[0:3] == 'chr':
+	vcf_chr = True
+sites_to_include = open(prune_in)
+sites_list = []
```
kachulis, line 82: make this a set not a list.  Since you will be checking whether sites are included over and over, you want the check for inclusion to be very fast, but don't care about being able to iterate over the set of included sites particularly efficienty.

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+    }
+    
+    output {
+    File map = "${output_prefix}.hapmap.txt"
+  }
+}
+
+workflow BuildHapMap {
+  File input_vcf
+  File header_file
```
kachulis, line 130: This is just a sequence dictionary.

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+  String intermediates_prefix
+  String output_prefix
+  Int prune_window
```
kachulis, line 133: set some reasonable defaults

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+    
+    output {
+    File map = "${output_prefix}.hapmap.txt"
+  }
+}
+
+workflow BuildHapMap {
+  File input_vcf
+  File header_file
+  String intermediates_prefix
```
kachulis, line 131: fine to have an option to change this, but definitely have a default so it's optional

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
+# BuildHaplotypeMap
+#
+# This workflow creates a haplotype map file containing a set of low LD sites for fingerprinting
+#
+# Inputs:
+#
+# 1. Genotyped multisample GVCF
+# 2. Header file. Standard SAM header (should match header of files to be fingerprinted)
```
kachulis, line 8: call this a sequence dictionary

kachulis (changes requested):
This is great @mollysacks, thanks!

A few comments and suggestions.  Also, some of the indentation is a bit weird, not sure if it's due to mixing tabs and spaces, or some fight between the github ui and ide.

HaplotypeMap/BuildHaplotypeMap.wdl
mollysacks: I think a VCF will work, I'm gonna try it just to check but I will change this if it does

HaplotypeMap/BuildHaplotypeMap.wdl
tmelman, line 26: @kachulis is this ok or should we be using dockerhub dockers?

@mollysacks make sure you have the dockerfile to this

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+      memory: "3500 MB"
+      docker: "docker.io/mollysacks/python:hapmap_builder"
+    }
+    
+    output {
+    File map = "${output_prefix}.hapmap.txt"
+  }
+}
+
+workflow BuildHapMap {
```
tmelman, line 128: add some comments. i.e. what does each step do?

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,167 @@
...
+  >>>
+  
+    runtime {
+      disks: "local-disk 1000 HDD"
+      memory: "3500 MB"
+      docker: "docker.io/mollysacks/python:hapmap_builder"
+    }
+    
+    output {
+    File map = "${output_prefix}.hapmap.txt"
```
tmelman, line 124: if this output_prefix is only used internally, it should probably be optional or set to some default.

HaplotypeMap/BuildHaplotypeMap.wdl
kachulis: as long as it's switched to the biocontainers one I think it's fine.  And then we don't need a Dockerfile, since it's externally maintained.

HaplotypeMap/BuildHaplotypeMap.wdl
mollysacks: Using bcftools is definitely easier! I'll add that in before I run the python code. Thanks!

tmelman (approved):
If @kachulis is satisfied I'm satisfied.

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -4,113 +4,129 @@
...
+# Format remaining variants into haplotype map format readable by Picard tools
+
 task reformat{
-	String output_prefix
-	File prune_in
-    File recode_vcf
-	File header_file
+  String output_prefix
+  File prune_in
+    File input_vcf
```
kachulis, line 93: indentation is odd here

HaplotypeMap/BuildHaplotypeMap.wdl
```diff
@@ -0,0 +1,190 @@
...
+    
+    output {
+    File map = "${output_prefix}.hapmap.txt"
+  }
+}
+
+workflow BuildHapMap {
+  File input_vcf
+  File sequence_dict
+  String intermediates_prefix
```
kachulis, line 147: this still isn't optional...

kachulis (changes requested):
Thanks for the changes @mollysacks I thinks this is almost there.  Just a couple small requests.  And add this workflow to the .dockstore.yml so that it can be imported into Terra via dockstore

kachulis (approved):
Looks good @mollysacks!

Issue #61, opened 2021-08-25
Handle missing APOL1 CKD risk alleles correctly


kachulis:
If a sample has a no-call in one of the APOl1 CKD risk alleles, we need to make sure the CKD adjustment script will still run succesfully and do something reasonable.

PR #62, opened 2021-08-25
Handle no-calls at APOl1 risk alleles in ckd adjustment

kachulis:
A no-call in the vcf will result in NA in the risk allele count file.  This pr then converts those NA's to 0, so treats no calls as 0 risk alleles.  

I haven't tested this yet at all, but I suspect the emerge 1-3 tests may hit this issue, so can be tested on that.

#61


tmelman (approved):
~~This is the test run in Terra. 

https://app.terra.bio/#workspaces/broadtagteam/TestPRSWrapper_copy_from_DSP/job_history/c31bb70e-2177-47cb-b589-b7ed5d4ca181

Good to go.~~

Issue #66, opened 2021-09-07
Keep sample names in output of FindSamplesAndBenchmark


meganshand:
Right now if you run "FindSamplesAndBenchmark" with multiple inputs, you lose which rows belong to which inputs in the final output summary.csv. We should take either the input sample name or even a user defined name and add it to the final summary output.

PR #68, opened 2021-09-16
Adds label to eval VCF in FindSamplesAndBenchmark and fixes call caching for Benchmark.wdl in Terra

meganshand:
Addresses #66 and also adds a dummy variable input to all of the tasks in Benchmark.wdl that don't take in  sample information. This helps fix the problem in Terra where tasks fail to call cache because it attempts to copy the outputs from a workspace you don't have access to. It's an optional input, so shouldn't make a difference for anyone not running in Terra (you can always just leave it empty).


BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -36,17 +37,25 @@ workflow FindSamplesAndBenchmark {
...
+    parameter_meta {
+      input_callset_labels: {description:"Labels for each input callset that is being evaluated. This will include this label in the output summary table under the 'Name' column."}
     }
 
     Int VCF_disk_size = ceil(size(input_callset, "GiB")) + 10
 
     call MakeStringMap as intervalsMap {input: keys=ground_truth_files, values=ground_truth_intervals}
     call MakeStringMap as lablesMap    {input: keys=ground_truth_files, values=truth_labels}
     call MakeStringMap as indexesMap   {input: keys=ground_truth_files, values=ground_truth_indexes}
+    call MakeStringMap as evalLabelsMap {input: keys=input_callset,     values = input_callset_labels}
```
rickymagner, line 53: Stylistic comment: the spaces around `=` should be removed here, so that the values are `values=input_callset_labels` to be consistent with previous lines.

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -163,7 +166,8 @@ workflow Benchmark {
                         refDict=refDict,
                         gatkTag=gatkTag,
                         subset_interval=CreateIntervalList.interval_list,
-                        preemptible=preemptible
+                        preemptible=preemptible,
+                        dummyInputForTerraCallCaching = dummyInputForTerraCallCaching
```
rickymagner, line 170: Stylistic comment: the spaces around `=` should be removed here, so that it says`dummyInputForTerraCallCaching=dummyInputForTerraCallCaching` to be consistent with previous lines. (Of course, this doesn't seem to be a consistent convention throughout the file but at least is held locally…)

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -185,7 +189,8 @@ workflow Benchmark {
             refDict=refDict,
             gatkTag=gatkTag,
             preemptible=preemptible,
-            subset_interval=CreateIntervalList.interval_list
+            subset_interval=CreateIntervalList.interval_list,
+            dummyInputForTerraCallCaching = dummyInputForTerraCallCaching
```
rickymagner, line 193: Stylistic comment: the spaces around `=` should be removed here, so that it says`dummyInputForTerraCallCaching=dummyInputForTerraCallCaching` to be consistent with previous lines.

rickymagner (approved):
Just made some stylistic comments based on surrounding text but otherwise looks OK.

PR #71, opened 2021-09-27
BenchmarkVCFs: Added input VCFs to disk size determination

### LINKED ISSUE #70: BenchmarkVCFs fails sometimes for large input files
The task `MatchEvalTruth` in BenchmarkVCFs sometimes runs out of disk space. This happens due to the task's disk size not being dependent on the input VCFs.

michaelgatzen:
- In task MatchEvalTruth the disk size was not dependent on the input VCFs. Now, both the VCFs and their indices are considered.
- Fixes #70


rickymagner (approved):
Looks like a reasonable thing to do.

PR #78, opened 2021-10-29
pulled out some parameters to be more easily accessed from terra

meganshand:


BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -134,10 +139,10 @@ workflow FindSamplesAndBenchmark {
                      jexlVariantSelectors = jexlVariantSelectors,
                      variantSelectorLabels = variantSelectorLabels,
                      referenceVersion = "1",
-                     doIndelLengthStratification=false,
+                     doIndelLengthStratification=doIndelLengthStratification,
                      gatkTag="4.0.11.0",
-                     requireMatchingGenotypes=true,
-                     passingOnly=true,
+                     requireMatchingGenotypes=requireMatchingGenotypes,
```
jamesemery, line 144: Should we also push the referenceVersion and GATK tags up?

jamesemery (approved):
Looks good. I remember those other two elements not actually being important but we should probably expose everything for the sake of it

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
meganshand: done

PR #65, opened 2021-08-27
Pulled out some Parameters from FindSamplesAndBenchmark.wdl to be accessible from the input.json

jamesemery:
``

PR #79, opened 2021-11-02
Accidentally added an extra parameter while rebasing a previous commit

meganshand:
This removes the extra copy of the parameter. This was introduced in #65


rickymagner (approved):
Just looked at the commit history for this file to see what happened, and this looks good to me!

PR #81, opened 2021-11-12
Added utility for collecting successful benchmarking outputs when some failed.

rickymagner:
See the README for a brief description of what this wdl does. I've found this useful for overcoming `FindSamplesAndBenchmark`'s tendency to fail on just a few samples when running a bunch at once, when I don't necessarily need every single output to do the analysis, just a good number of them. This is motivated by me trying to rerun the wdl multiple times to "fill in the failed tasks" using fast call caching, but then ending up failing some of the previously successful ones and not getting to the compilation final step of the wdl.


tmelman (approved):
This is fine, see suggestions.

Utilities/CollectBenchmarkSucceeded.wdl
```diff
@@ -0,0 +1,62 @@
...
+        String sub_id
+    }
+
+    command <<<
+        # Keep track of shard number in temp file to handle loss of scope in loop
+        echo 0 > number.txt
+
+        # Double loop allows compatibility with one or multiple workflow sample pairs
+        gsutil ls gs://~{ws_name}/~{sub_id}/FindSamplesAndBenchmark/ | while read -r workflow ; do
+            gsutil ls "${workflow}call-BenchmarkVCF" | while read -r shard ; do
```
tmelman, line 33: Suggest storing the list: 
`shard_list=($(gsutil ls "${workflow}call-BenchmarkVCF))` and then using a for loop for each shard in the list.

rickymagner:
Made some changes as suggested to remove need for pipes, which causes some variable scoping issues. Hopefully it looks more readable now!

Issue #77, opened 2021-10-29
Example inputs for FindSamplesAndBenchmark have the wrong picard jar


meganshand:
This jar https://github.com/broadinstitute/palantir-workflows/blob/main/BenchmarkVCFs/FindSamplesAndBenchmark_input.json#L10 

doesn't work as expected. We need a cloud friendly Picard jar, but apparently the latest version of picard gives an error: `Exception in thread "main" java.nio.file.ProviderNotFoundException: Provider "gs" not found`. We should track this down so that a released version of Picard can be used for fingerprinting in the FindSamlesAndBenchmark WDL.

rickymagner:
This should be fixed in #86

Issue #75, opened 2021-10-25
Truth data default "requester pays" buckets for FindSamplesAndBenchmark aren't accessible in Cromwell through Terra


rickymagner:
When using the default inputs to `FindSamplesAndBenchmark` on Terra, Cromwell complains about not being able to access the requester-pays buckets housing the GIAB truth data. It seems a reasonable fix would be modifying the wdl to allow the user to input a specific project to bill at this step, e.g. like in [this](https://github.com/broadinstitute/gatk/blob/master/scripts/mitochondria_m2_wdl/MitochondriaPipeline.wdl#L229) example.

rickymagner:
Fixed in #86

Issue #82, opened 2021-11-16
Not specifying optional stratIntervals/stratLabels in FindSamplesAndBenchmark results in error.


samuelklee:
Trying to do this on Terra gave me this error at the BenchmarkVCFs step: `Failed to evaluate input 'stratLabels' (reason 1 of 1): assertion failed: base member type WomMaybeEmptyArrayType(WomStringType) and womtype WomMaybeEmptyArrayType(WomSingleFileType) are not compatible`. Haven't looked too closely, but perhaps the following line is to blame? https://github.com/broadinstitute/palantir-workflows/blob/92b16f58f2a169f8982ef6a7f25a59cc99f6fdf8/BenchmarkVCFs/FindSamplesAndBenchmark.wdl#L109

There might also be some interaction between how the optional array types are specified at the FindSamplesAndBenchmark and BenchmarkVCFs levels, but I must admit I'm not up to date on best practices for optionals in WDL.

meganshand:
Can you provide your inputs? Or were you using the inputs json here: https://github.com/broadinstitute/palantir-workflows/blob/main/BenchmarkVCFs/FindSamplesAndBenchmark_input.json ?

samuelklee:
I think you should be able to see the job at https://job-manager.dsde-prod.broadinstitute.org/jobs/27ac8a1e-777d-4478-bc07-9e42d38197c4. Let me know if anyone else would like access.

meganshand:
Ah thanks I see that the stratInterval + stratLabel are empty not optional, and now that I reread the title of this issue I understand :)

michaelgatzen:
Can I get access to the workflow as well please?

samuelklee:
@michaelgatzen you should have reader access to the workspace and I've made the workflow public (although I think it matches the current version in this repo).

rickymagner:
This should be fixed in #86. If one uses the wdl from the main branch, it should work with this use case! Let us know if there are any other problems

PR #86, opened 2021-12-01
Fixing some issues and updating FindSamplesAndBenchmark

rickymagner:
This PR makes a number of changes to `FindSamplesAndBenchmark`. These changes were inspired by a few recent issues with the wdl, most of which should be fixed now. We break down the changes into a few categories.

## Cosmetic Changes
* A few typos were fixed.
* Some whitespaces were normalized.
* Some default values were put directly in the wdl for less commonly changed inputs. 
* The `annotation_intervals` input was also never used so it was removed, and the `ref_fasta_sdf` didn't seem to have any significance as it only was used in allocating disk size, so this input was also removed. 

## GATK/Picard/bcftools Version Control
* Updated the dockers for the GATK tasks to use whichever tag the user inputs. The default input has been changed to the latest official tag as of the time of this PR. This set-up is consistent with e.g. `BenchmarkVCFs`, and necessary for staying modern as the GITC docker is officially retired in favor of app-specific dockers. 
* Updated the default picard jar in the inputs json to be of the latest version, stored in a neutral location (`broad-dsde-methods`). This addresses and resolves issue #77. 
* There is now an entry point for a `picardDocker` input. It would be nice if this were a minimal official docker with just Picard install, but for now it defaults to a GATK docker which has all the Java ingredients needed to run a Picard jar. 
* There is also an input for the `bcftoolsDocker` used for the tasks which previously had a hardcoded docker image for `bcftools`. 

## Rewriting `CrosscheckFingerprints` and `PickMatches` tasks
This section is meant to resolve issue #75. The problem was that `CrosscheckFingerprints` would use localized file paths in its outputs, which resulted in `PickMatches` being unable to find the appropriate file paths. (I don't fully understand why this issue was avoided in the past by the code…) This was particularly an issue when using the default truth inputs with requester-pays buckets, where it becomes necessary for Cromwell to localize the files to access them. Here are the changes made that should fix all of these problems.
* `CrosscheckFingerprints` now outputs two additional short files: the global bucket file paths to the truth files, and the localized Cromwell paths for the files which ultimately get printed to the Picard output file. These are necessary to pass along to `PickMatches`.
* The parameter_meta for the ground_truth_files input has `localization_optional` set to `false` now, to ensure the files get localized to get the correct behavior when accessing requester-pays buckets for the truth data.
* `PickMatches` has been fully rewritten in Python to add readability and perform more complex text manipulations than the previous `awk` script. This task now takes the output of `CrosscheckFingerprints` and manually replaces the localized Cromwell file paths with the appropriate global bucket paths, using the lists from that task. 
* Previously the truth labels input was never put into the BenchmarkVCFs subtask. It was not possible to modify directly by using the label corresponding to the truth vcf in the labels map because the file path was localized, as mentioned above. This is now fixed, and the user defined truth labels are used in the final output csv. 

It should be noted that if the input files are also from requester-pays buckets, it's possible similar issues may arise, but this hasn't been tested. Regardless, things should work now at least with our default requester-pays truth buckets, and user provided vcfs.


michaelgatzen:
Also added a commit to fix #82 (failing when not specifying stratLabels or stratIntervals) and fixed an error in the default input JSON

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -2,86 +2,86 @@ version 1.0
...
+        Array[String]? annotationNames = []
+
         File ref_fasta
         File ref_fasta_index
         File ref_fasta_dict
-        File ref_fasta_sdf
 
-        File haplotype_database 
-        File picard_cloud_jar
+        File haplotype_database = "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.haplotype_database.txt"
```
michaelgatzen, line 21: I would say either default all reference files to hg38 or none of them, in case users want to run it in hg19

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -211,98 +211,110 @@ struct Match{
...
+        CODE
     >>>
+
     output {
         Array[Array[String]] matches = read_tsv("matches.tsv")
     }
+
     runtime {
-        docker: "ubuntu"
+        docker: "us.gcr.io/broad-dsde-methods/rm-python-data-slim"
```
michaelgatzen, line 311: Is there a more "official" docker image than a personalized one? This is probably the most efficient docker image, but it's good to either point to a well-maintained docker image (with the Dockerfile publicly available) or include the Dockerfile in this repo if we want to have a custom one

michaelgatzen: Usually I don't think we optimize for docker image size that much, so if there's a public one that's larger in size then that's fine

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
rickymagner: Good point. I added hg38 for the default reference.

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
rickymagner: Unfortunately I haven't been able to find even an "official" looking docker just with pandas installed! I tried looking for generic "data" dockers that would include those libraries as well, but was surprised to not find anything obvious. If you know of any I'd be happy to substitute it.

rickymagner: Alternatively I could rename it so it looks less "personalized." The dockerfile exists here in palantir-workflows, so it could just be a docker image we "officially" maintain, though I agree I don't like the idea of using something so generic that it should be standard but doesn't seem to be provided by anyone closer to the developers as far as I could tell.

michaelgatzen: Oh awesome, I didn't realize the Dockerfile was in this repo already, so yeah, we should rename it to have a name without the rm, and perhaps put it into a requester pays bucket, so that we don't get charged every time someone pulls the docker

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -14,15 +14,16 @@ workflow FindSamplesAndBenchmark {
...
-        File ref_fasta
-        File ref_fasta_index
-        File ref_fasta_dict
-
+        File ref_fasta = "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta"
+        File ref_fasta_index = "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta.fai"
+        File ref_fasta_dict = "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dict"
         File haplotype_database = "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.haplotype_database.txt"
-        File picard_cloud_jar = "gs://broad-dsde-methods/picard/picardcloud-2.26.6.jar"
+        String referenceVersion = "1"
```
michaelgatzen, line 21: I think this should be "hg38", per the documentation of BenchmarkVCF

rickymagner: Good call

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
rickymagner: Did the renaming in the latest commit. I'm not sure how to turn a docker image to requester pays. It might be possible if it's stored in a bucket somewhere, by setting it for the bucket, but I haven't been able to hunt down where it's stored yet.

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -211,98 +210,110 @@ struct Match{
...
-            } 
-        $(f["RESULT"])!~/MISMATCH/ && $(f["RESULT"])~/MATCH/ {
-            print $(f["LEFT_FILE"]), $(f["LEFT_SAMPLE"]), $(f["RIGHT_FILE"]), $(f["RIGHT_SAMPLE"])
-            } 
-        AWK
-
-
-        awk -F'\t' -f prog.awk ~{crosscheck_results} > matches.tsv
+        # Import crosscheck results into DataFrame
+        crosscheck = pd.read_csv("~{crosscheck_results}", sep='\t', skiprows=6)
```
gileshall, line 287: `skiprows=6` seems fragile to me, and not easily explainable to someone who was new to the code.  I don't know why you are skipping the first six rows, but I suggest calculating this skip based on something intrinsic.  It might never change, but it at least documents your intent with code.

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
```diff
@@ -350,59 +362,60 @@ task MakeStringMap {
...
+        -O ~{output_basename}.vcf.gz \
+        --exclude-non-variants \
+        --select-type-to-exclude SYMBOLIC
+    >>>
+    runtime {
+        preemptible: preemptible_tries
+        memory: "12 GB"
+        cpu: "1"
+        disks: "local-disk " + ceil(disk_size) + " HDD"
+        docker: "us.gcr.io/broad-gatk/gatk:"+gatkTag
```
gileshall, line 410: Just a suggestion, but in some places, when the docker string can change, it is either calculated like the above in the task -or- it's just a variable.  I know this idiom is carried over, but I think it would be great if there was a paragraph at the top that calculated and finalized all docker strings such that task level `docker:` just referred to appropriate docker variable.

gileshall (changes requested):

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
rickymagner: I agree this is a magic number here which would be better to avoid. The reason for it is to skip a sam header generated from the crosscheck fingerprints task. I'll try to think of a better way to handle this.

rickymagner: I settled on exploiting the fact that picard outputs use `#` to preface header/comment lines, so check out the new `read_csv` line which I think handles this perfectly without any magic numbers.

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
rickymagner: I put some docker variables in globally in the wdl, so check out the new convention. Thanks for this idea!

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
gileshall: nailed it

BenchmarkVCFs/FindSamplesAndBenchmark.wdl
gileshall: this looks so much cleaner, nice work!

gileshall (approved):

PR #88, opened 2021-12-16
PRS Reportable Range Check

kachulis:
In PRSWrapper, If PRS z-scores is outside defined reportable range, results are changed to "NOT_RESULTED" and reason is recorded in "~{condition}_not_resulted_reason" column.


kachulis:
@edytamalolepsza this is for only outputting if z-score is within reportable range.  I have tested it myself, but can you also try it out and make sure it behaves as you would expect?  (you can set the reportable range to be small (like 1.5, for example) to get samples to fall outside of the range.

michaelgatzen (approved):
Looks great!

PR #90, opened 2022-01-24
Add Interaction terms to PRS pipeline 

kachulis:
Adds interaction terms to the PRS pipeline.  For a listed interaction, each allele can only participate in the interactions once.  So, if a sample is het in one interaction allele, and hom in the other, it will have the 1x interaction weight added to the score.  If it is hom in both interaction alleles, it will have 2x interaction weight added.  And if it is het in one and doesn't contain the other, will not have the interaction weight added.  Additionally, if the two interaction alleles in the interaction are the same, then the interaction weight will be added if and only if the sample is hom in the interaction allele.

There is also a bit of a hack included here using the `SelfExclusiveSites`.  This is a set of alleles of which only a specified number are allowed to be included, or the interaction term is not added to the score.  This is a roundabout way to implement the algorithm described in the [T1D GRS2 paper ](https://pubmed.ncbi.nlm.nih.gov/30655379/) in a flexible and reusable manner.

### DIFF
--- a/ImputationPipeline/ScoringPart.wdl
+++ b/ImputationPipeline/ScoringPart.wdl
@@ -312,7 +312,7 @@ task AddInteractionTermsToScore {
 
 					add_allele_to_count(site_1, allele_1, interactions_allele_counts)
 					add_allele_to_count(site_2, allele_2, interactions_allele_counts)
-					interaction_idct[(site_1, allele_1, site_2, allele_2)] = weight
+					interaction_dict[(site_1, allele_1, site_2, allele_2)] = weight
 					positions.add((chrom_1, pos_1))
 					positions.add((chrom_2, pos_2))
 
--- a/ImputationPipeline/ScoringPart.wdl
+++ b/ImputationPipeline/ScoringPart.wdl
@@ -391,7 +391,7 @@ task AddInteractionTermsToScore {
 
 		#add interaction scores to linear scores
 		df_interaction_score = pd.DataFrame({"sample_id":samples, "interaction_score":interaction_scores}).set_index("sample_id")
-		df_scores=pd.read_csv("~{scores}", sep="\t").astype({'#IID#':'string'}).set_index("#IID")
+		df_scores=pd.read_csv("~{scores}", sep="\t").astype({'#IID':'string'}).set_index("#IID")
 		df_scores = df_scores.join(df_interaction_score)
 		df_scores['SCORE1_SUM'] = df_scores['SCORE1_SUM'] + df_scores['interaction_score']
 		df_scores.to_csv("~{basename}_scores_with_interactions.tsv", sep="\t")
--- a/ImputationPipeline/ScoringPart.wdl
+++ b/ImputationPipeline/ScoringPart.wdl
@@ -1,10 +1,14 @@
 version 1.0
 
+import "Structs.wdl"
+
 workflow ScoringImputedDataset {
 	input { 
 	File weights # disease weights file. Becauase we use variant IDs with sorted alleles, there is a task at the bottom of this workflow
 	#  that will allow you to sort the variants in this weights file (`SortWeights`)
 	File? interaction_weights # wieghts for snp interactions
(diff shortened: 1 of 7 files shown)


ImputationPipeline/imputation_interaction_python/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# default ubuntu version: 20.04
+ARG UBUNTU_VERSION=20.04
+FROM ubuntu:${UBUNTU_VERSION}
+
+# the following argument is needed for pkg-config to
+# install without hanging on user input
+ARG DEBIAN_FRONTEND=noninteractive
```
michaelgatzen, line 7: The Dockerfile compiles without these lines for me, do we need them?

ImputationPipeline/ScoringPart.wdl
```diff
@@ -229,6 +254,156 @@ task ScoreVcf {
...
+			else:
+				current_end = site[1] + 1
+
+		#last block
+		blocks_to_read.append(current_chrom + ":" + str(current_start) + "-" + str(current_end))
+
+		#count interaction alleles for each sample
+		for block in blocks_to_read:
+			for variant in vcf(block):
+				alleles = [a for a_l in [[variant.REF], variant.ALT] for a in a_l]
```
michaelgatzen, line 356: I feel like there must be a more comprehensible way to write this section, but since it does what it's supposed to do let's leave it!

ImputationPipeline/ScoringPart.wdl
```diff
@@ -229,6 +254,156 @@ task ScoreVcf {
...
+		if ~{if defined(sites) then "True" else "False"}:
+			with open("~{sites}") as f_sites:
+				sites = {s.strip() for s in f_sites}
+		else:
+			sites = {}
+		with open("~{interaction_weights}") as f:
+			for line in f:
+				if line.strip():
+					line_split = line.split()
+					if weight := read_as_float(line_split[8]):
```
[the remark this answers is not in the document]
michaelgatzen: (if this was added because of the header line, the DictReader takes care of that)

michaelgatzen (approved):
I'll also approve it now because none of the things I commented on above are _wrong_, that way you can merge it whenever you want, but we can of course do another round of review if you do make changes

ImputationPipeline/imputation_interaction_python/Dockerfile
kachulis: nope!

ImputationPipeline/ScoringPart.wdl
```diff
@@ -229,6 +254,156 @@ task ScoreVcf {
...
+
+		interactions_allele_counts = dict()
+		interactions_dict = dict()
+		positions = set()
+		if ~{if defined(sites) then "True" else "False"}:
+			with open("~{sites}") as f_sites:
+				sites = {s.strip() for s in f_sites}
+		else:
+			sites = {}
+		with open("~{interaction_weights}") as f:
```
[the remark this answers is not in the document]
kachulis: ha, yes that is a much better way to do this

kachulis:
@michaelgatzen took you suggestion about using csv.  The results are identical to before the change, so I'm pretty confident I didn't mess it up, but if you could do a quick once over before I merge this that'd be great.

ImputationPipeline/ScoringPart.wdl
```diff
@@ -230,6 +255,160 @@ task ScoreVcf {
...
+			for line in csv.DictReader(f, delimiter='\t'):
+				site_1 = line['id_1']
+				site_2 = line['id_2']
+				if len(sites) == 0 or site_1 in sites and site_2 in sites:
```
michaelgatzen, line 304: I always like to set parentheses when having an "a or b and c" condition for clarity, because different languages sometimes process them differently (I think), but feel free to leave it for this merge, we can always work on aesthetics later

michaelgatzen (approved):
The changes look good, there's one minor comment that you can ignore if you don't want to do all the testing again

PR #92, opened 2022-02-01
Aggregate PRS Output

michaelgatzen:
- PRSWrapper.wdl
  - Added columns `batch_id` and `is_control` to results.csv
- New file: AggregatePRSResults
  - Aggregate multiple results CSV files into summarized tables
    - Check that all `batch_id`s are the same
    - Check that there is exactly one control sample within every batch
    - Extract control sample
    - Extract results in two shapes: `all_results` and `summarized_results`
    - Extract score distribution
  - Create population PCA plot
  - Build HTML report
    - Compare control sample to expected control result
    - Batch summary
    - Score distribution
    - Population PCA
    - Individual results
- Added to Dockstore.yml
- Fixed bug in ScoringPart.wdl

### DIFF
--- a/ImputationPipeline/AggregatePRSResults.wdl
+++ b/ImputationPipeline/AggregatePRSResults.wdl
@@ -164,7 +164,6 @@ task BuildHTMLReport {
     library(dplyr)
     library(stringr)
     library(kableExtra)
-    knitr::opts_chunk$set(echo = TRUE)
     batch_results <- read_tsv("~{batch_results}")
     batch_summary <- read_tsv("~{batch_summarised_results}")
     batch_summary <- batch_summary %>% rename_with(.cols = -condition, ~ str_to_title(gsub("_"," ", .x)))
--- a/ImputationPipeline/AggregatePRSResults.wdl
+++ b/ImputationPipeline/AggregatePRSResults.wdl
@@ -101,12 +101,11 @@ task PlotPCA {
     library(purrr)
     library(ggplot2)
 
-    target_pcs <- c("~{sep='","' target_pc_projections}") %>% map(read_tsv) %>% reduce(bind_rows) %>% mutate(cohort="~{batch_id}", size=0.5, alpha=1)
-    population_pcs <- read_tsv("~{population_pc_projections}") %>% mutate(cohort="~{population_name}", size=0.1, alpha=0.1)
+    target_pcs <- c("~{sep='","' target_pc_projections}") %>% map(read_tsv) %>% reduce(bind_rows))
+    population_pcs <- read_tsv("~{population_pc_projections}"))
 
-    all_pcs <- bind_rows(target_pcs, population_pcs)
-    ggplot(all_pcs, aes(x=PC1, y=PC2)) +
-      geom_point(aes(color=cohort, size=size, alpha=alpha)) +
+    ggplot(population_pcs, aes(x=PC1, y=PC2), size=0.1, alpha=0.1, , color="~{population_name}") +
+      geom_point(data=target_pcs, color="~{batch_id}") +
       theme_bw()
 
     ggsave(filename = "~{batch_id}_PCA_plot.png", dpi=300, width = 6, height = 6)
--- a/ImputationPipeline/AggregatePRSResults.wdl
+++ b/ImputationPipeline/AggregatePRSResults.wdl
@@ -197,9 +197,9 @@ task BuildHTMLReport {
 
 
     ## Control Sample
-    ```{r control, echo = FALSE, results = "asis"}
+    \`\`\`{r control, echo = FALSE, results = "asis"}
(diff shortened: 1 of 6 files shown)


ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -0,0 +1,235 @@
...
+
+    if (num_batches != 1) {
+      stop(paste0("There are ", num_batches, " batch IDs in the input tables, however, only 1 is expected."))
+    }
+
+    batch_id <- results$batch_id[1]
+
+    num_control_samples <- sum(results$is_control=="true")
+
+    if (num_batches != 1) {
```
kachulis, line 72: `num_batches` should be `num_controls`

ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -0,0 +1,235 @@
...
+      stop(paste("There are ", num_control_samples, " control samples in the input tables, however, only 1 is expected."))
+    }
+
+    write_tsv(results, paste0(batch_id, "_all_results.tsv"))
+
+    write_tsv(results %>% filter(is_control == "true"), paste0(batch_id, "_control_results.tsv"))
```
kachulis, line 78: should be able to just do `filter(is_control)`.  "true" should be parsed as a logical by read_tsv.

ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -0,0 +1,235 @@
...
+
+    results_pivoted <- results %>% select(-batch_id) %>% select(-is_control) %>% pivot_longer(!sample_id, names_to=c("condition",".value"), names_pattern="([^_]+)_(.+)")
```
kachulis, line 80: use a single select `select(-batch_id, -is_control)`

ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -0,0 +1,235 @@
...
+
+    ## Batch Score distribution
+    ![](~{score_distribution})
+
+    ## PCA
+    ![](~{pc_plot})
+
+    ## Individual Sample Results
+    \`\`\`{r sample results , echo = FALSE, results = "asis"}
+    kable(batch_all_results %>% mutate(across(ends_with("risk"), ~ kableExtra::cell_spec(.x, color=ifelse(is.na(.x), "blue", ifelse(.x=="NOT_RESULTED", "red", ifelse(.x == "HIGH", "orange", "green")))))), digits = 2)
```
kachulis, line 219: I think we should filter out the control sample from batch_all_results before displaying the table

kachulis (changes requested):
Looks great @michaelgatzen!  some minor comments about using more dplyr.

michaelgatzen:
Addressed all comments and made a few changes to the HTML report:
- Added column to the control table describing which one is the batch and which one is the control sample
- Removed `is_control` column from the Individual Sample Results table

ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -198,7 +200,7 @@ task BuildHTMLReport {
 
     ## Control Sample
     \`\`\`{r control, echo = FALSE, results = "asis"}
-    kable(list(batch_control_results, expected_control_results) %>% reduce(bind_rows) %>% select(-batch_id) %>% select(-is_control) , digits = 2)
+    kable(list(batch_control_results, expected_control_results) %>% reduce(bind_rows) %>% select(-batch_id, -is_control) %>% add_column(Sample=c('Batch control', 'Expected control'), .before='sample_id'), digits = 2)
```
kachulis, line 203: I think we should drop the sample_id column as well here.  All we care about is which is the batch control, and which is the expected control, the sample id of the control is unimportant.

ImputationPipeline/AggregatePRSResults.wdl
```diff
@@ -57,27 +57,27 @@ task AggregateResults {
...
+    batch_ids <- results %>% pull(batch_id) %>% unique()
 
-    if (num_batches != 1) {
-      stop(paste0("There are ", num_batches, " batch IDs in the input tables, however, only 1 is expected."))
+    if (length(batch_ids) != 1) {
+      stop(paste0("There are ", length(batch_ids), " batch IDs in the input tables, however, only 1 is expected."))
     }
 
-    batch_id <- results$batch_id[1]
+    batch_id <- batch_ids[1]
```
kachulis, line 68: you don't actually need to select the particular element here.  In R a single value is actually vector of length 1.  So, since batch_ids is guaranteed to be length 1 a this point in the code, `batch_ids`, `batch_ids[1]`, and `batch_ids[[1]]` are all the same.

kachulis (changes requested):
thanks @michaelgatzen , just two very small additional requests.

ImputationPipeline/AggregatePRSResults.wdl
michaelgatzen: Wizardry!

kachulis (approved):

PR #93, opened 2022-02-06
Notebook for manual QC editing of prs results table

kachulis:


ImputationPipeline/notebooks/CreateManuallyQCdPRSResultsFile.ipynb
```diff
@@ -0,0 +1,859 @@
...
+    "        self.finished_sample_failures_button.close()\n",
+    "        self.build_fail_condition_section()\n",
+    "        self.printFailedConditionsAllSamples()\n",
+    "\n",
+    "\n",
+    "\n",
+    "\n",
+    "    def build_fail_condition_section(self):\n",
+    "        self.out_failed_condition_all_samples = widgets.Output(layout={'border': '1px solid black'})\n",
+    "        self.condition_failure_selection = widgets.Dropdown(options = [None] + self.conditions,\n",
```
michaelgatzen, line 595: One thing about the `None` option, if someone would select a condition and then go back to the empty option, they would still be able to enter a reason, and if they then click "Fail" the notebook crashes. A quick fix in `clickFailForAllSamplesButton` if the `None` option is selected will fix that, but that's also a very unlikely scenario in my opinion

michaelgatzen (approved):
I commented on the only (not very realistic) bug I could find by looking through the code, but the rest looks great!

ImputationPipeline/notebooks/CreateManuallyQCdPRSResultsFile.ipynb
kachulis: good find!

PR #95, opened 2022-02-10
Changed batch_id to lab_batch and is_control to is_control_sample

michaelgatzen:
- Also checked in CreateSampleSets notebook
  - Discarded `batch_` prefix for sample_sets
  - Changed timestamp to Eastern time instead of UTC
  - Automatically detecting workspace information


kachulis (approved):
looks great, thanks @michaelgatzen

PR #97, opened 2022-02-14
FE: rewrote F1 evaluation using Pandas

michaelgatzen:
- New (more correct) calculation of the F1 scores at each sampling position (this solves the issue of "gaps" in the plots)
- Automatic detection of the stratifiers

### DIFF
--- a/FunctionalEquivalence/subworkflows/F1Evaluation.wdl
+++ b/FunctionalEquivalence/subworkflows/F1Evaluation.wdl
@@ -5,7 +5,6 @@ workflow F1Evaluation {
         Array[File] roc_tables
         String tool1_label
         String tool2_label
-        Array[String] stratifiers
         String? additional_label
         Boolean signed_difference = false
         Int? mem_gb
@@ -17,7 +16,6 @@ workflow F1Evaluation {
             roc_tables = roc_tables,
             tool1_label = tool1_label,
             tool2_label = tool2_label,
-            stratifiers = stratifiers,
             additional_label = additional_label,
             signed_difference = signed_difference,
             mem_gb = mem_gb,
@@ -37,7 +35,6 @@ task F1EvaluationTask {
         Array[File] roc_tables
         String tool1_label
         String tool2_label
-        Array[String] stratifiers
         String? additional_label
         Boolean signed_difference = false
         Int? mem_gb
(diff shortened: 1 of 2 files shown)


FunctionalEquivalence/subworkflows/F1Evaluation.wdl
```diff
@@ -107,126 +85,131 @@ def read_file(data, filename, stratifiers):
...
-def extract_data_from_dict(score_range, data, dataset, tool, region, var_type):
-    return [
-                np.array([
-                    data[dataset][replicate][tool][region][var_type][score] if score in data[dataset][replicate][tool][region][var_type] else np.nan
-                    for replicate in data[dataset].keys()
-                ])
-                for score in score_range
-            ]
-    
+    file_data = pd.read_csv(filename, header=6, delimiter='\t')
```
rickymagner, line 88: The magic 6 here looks like Picard files to me. Regardless, I think if you're trying to skip some headers/comments at the top, try using the `comment` input field, e.g. `comment='#'` to automatically skip any lines beginning with a `#` in favor of the hard-coded number of lines to skip.

FunctionalEquivalence/subworkflows/F1Evaluation.wdl
```diff
@@ -235,52 +218,55 @@ def diff_plot(ax, data, dataset, var_type, region, tool1_label, tool2_label, sig
     else:
         print('Invalid plot data for {} {} {}'.format(dataset, var_type, region))
```
rickymagner, line 219: This and the other `.format`'s could probably be made for readable using f-strings, if you want ("while you're here"). 

E.g. At this line, you can use
`print(f'Invalid plot data for {dataset} {var_type} {region}')`

This is particularly relevant for some of the longer strings in the new code below.

rickymagner, line 219: This and the other `.format`'s could probably be made for readable using f-strings, if you want ("while you're here"). 

E.g. At this line, you can use
`print(f'Invalid plot data for {dataset} {var_type} {region}')`

FunctionalEquivalence/subworkflows/F1Evaluation.wdl
```diff
@@ -299,7 +285,7 @@ def plot_aggregated_data_compact(data, stratifiers, dataset, tool1_label, tool2_
     fig.suptitle('Dataset: {}'.format(dataset) + ('' if not additional_label else ', {}'.format(additional_label)) + '\n' +
                 ('Signed ' if signed_difference else 'Absolute ') +
                 r'$F_1 = \frac{TP}{TP + \frac{1}{2} (FP + FN)}$ score differences for calls with $QUAL \geq q$' +
-                ' (# replicates: {})'.format(len(data[dataset])))
+                ' (# replicates: {})'.format(len(data.loc[data['dataset'] == dataset].replicate.unique())))
```
rickymagner, line 288: This figure title in particular can be made more readable with some f-strings (see last comment).

rickymagner, line 288: Here in particular can be made more readable with an f-string (see last comment).

rickymagner (approved):
Looks good overall. See a few small comments for some suggested improvements in readability.

FunctionalEquivalence/subworkflows/F1Evaluation.wdl
michaelgatzen: Unfortunately this doesn't work because annoyingly, the header line also starts with a `#` in these RTG vcfeval files. They look like this:
```
#comment
...
#comment
#score	true_positives_baseline ...
value2	value2
```

FunctionalEquivalence/subworkflows/F1Evaluation.wdl
michaelgatzen: Changed all format strings to f-strings except for one

PR #99, opened 2022-03-02
PRS missing sites QC

kachulis:
Checks if any sites used for training of ancestry model are missing.  If there are missing sites, calculates possible shift due to missing sites.  Include missing site info in aggregated qc report.

### DIFF
--- a/ImputationPipeline/PRSWrapper.wdl
+++ b/ImputationPipeline/PRSWrapper.wdl
@@ -80,7 +80,7 @@ workflow PRSWrapper {
   call CombineMissingSitesShiftedScores {
     input:
       missing_sites_shifted = missing_sites_shifted,
-      lab_batch = lab_batch
+      lab_batch = lab_batch_id
   }
 
 
--- a/ImputationPipeline/PRSWrapper.wdl
+++ b/ImputationPipeline/PRSWrapper.wdl
@@ -75,10 +75,20 @@ workflow PRSWrapper {
       is_control_sample = is_control_sample
   }
 
+  Array[File] missing_sites_shifted = select_all(ScoringImputedDataset.missing_sites_shifted_scores)
+
+  call CombineMissingSitesShiftedScores {
+    input:
+      missing_sites_shifted = missing_sites_shifted,
+      lab_batch = lab_batch
+  }
+
+
   output {
     File results = JoinResults.results
     File pcs = select_first(ScoringImputedDataset.pc_projection)
     String lab_batch = lab_batch_id
+    File missing_sites_shifts = CombineMissingSitesShiftedScores.missing_sites_shifts
   }
 }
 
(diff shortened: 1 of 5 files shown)


ImputationPipeline/PRSWrapper.wdl
```diff
@@ -75,10 +75,20 @@ workflow PRSWrapper {
       is_control_sample = is_control_sample
   }
 
+  Array[File] missing_sites_shifted = select_all(ScoringImputedDataset.missing_sites_shifted_scores)
+
+  call CombineMissingSitesShiftedScores {
+    input:
+      missing_sites_shifted = missing_sites_shifted,
+      lab_batch = lab_batch
```
michaelgatzen, line 83: I have no idea how this works, because it shouldn't know `lab_batch` at this point. 
```suggestion
      lab_batch = lab_batch_id
```

ImputationPipeline/ScoringTasks.wdl
```diff
@@ -199,6 +199,123 @@ task AddInteractionTermsToScore {
...
+        if id in missing_sites:
+          missing_site_weight = float(line[2])
+          if missing_site_weight > 0:
+            max_error_up += 2 * missing_site_weight
+          else:
+            max_error_down += 2 * missing_site_weight
+
+
+    if ~{if defined(weight_set.interaction_weights) then "True" else "False"}:
+      with open("~{if (defined(weight_set.interaction_weights)) then select_first([weight_set.interaction_weights]) else ''}") as f_interaction_weights:
```
michaelgatzen, line 286: Same here

ImputationPipeline/ScoringTasks.wdl
```diff
@@ -199,6 +199,123 @@ task AddInteractionTermsToScore {
...
+
+    sites_in_weight_set = set()
+    with open("~{weight_set.linear_weights}") as f_linear_weights:
+      linear_weights_reader = csv.reader(f_linear_weights, delimiter='\t')
+      next(linear_weights_reader)
+      for line in linear_weights_reader:
+        sites_in_weight_set.add(line[0])
+
+    if ~{if (defined(weight_set.interaction_weights)) then "True" else "False"}:
+      with open("~{if (defined(weight_set.interaction_weights)) then select_first([weight_set.interaction_weights]) else ''}") as f_interaction_weights:
```
michaelgatzen, line 224: You can probably skip the second `if defined here`, can't you?

ImputationPipeline/PRSWrapper.wdl
kachulis: Looks like an error during merging for a rebase after I had tested it.  Guess I should retest :)

ImputationPipeline/ScoringTasks.wdl
kachulis: You would think, but the wdl spec calls for a runtime workflow failure is a select_first is evaluated  with no defined inputs.  However, this entire ugly if/then/select_first clause can just be replaced by `select_first([weight_set.interaction_weights, ''])`, which is definitely what I should have done in the first place, lol

kachulis: what am I talking about, I don't even need the selec_first, it will just evaluate to an empty string if it's not defined anyway.  oy vey, long week

michaelgatzen (approved):

PR #111, opened 2022-08-25
Updated python-data-slim and using it in FunctionalEquivalence WDLs

michaelgatzen:
Fixes bug where pandas was not installed in the FE evaluation dockers

### DIFF
--- a/FunctionalEquivalence/FunctionalEquivalence.wdl
+++ b/FunctionalEquivalence/FunctionalEquivalence.wdl
@@ -383,8 +383,6 @@ task CreateHTMLReport {
 
     command <<<
         set -xeuo pipefail
-        
-        source activate fe_evaluation
 
         fe_plots_base64=$(base64 -w 0 ~{merged_fe_plots})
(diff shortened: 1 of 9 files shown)


FunctionalEquivalence/FunctionalEquivalence.wdl
```diff
@@ -444,7 +442,7 @@ EOF
     >>>
 
     runtime {
-        docker: "us.gcr.io/broad-dsde-methods/functionalequivalence/fe_evaluation:1.0.0"
+        docker: "us.gcr.io/broad-dsde-methods/mgatzen/python-data-slim:98f9877658fbb4a8298fbd69f7daf0fa5415c6f79d7f4d0ef47a4ed41a5c5ad3"
```
rickymagner, line 445: Why does this have a weird hash-looking version number? Could we use a more human-readable version, especially given you've bumped the dependency versions? It'd be nice if this version number was loosely correlated with those.

michaelgatzen:
@rickymagner Would you mind reviewing this again so that I can merge it ASAP? It fixes an issue that we need for a Terra workshop on Friday

PR #112, opened 2022-09-10
Yf add combine summaries wf

yfarjoun:
- Combine summaries
- fix 'miniwdl check' complaints


rickymagner:
Hi Yossi, just finished looking at this. Although I'm still confused how the empty file gets away with it (as we chatted about), I think these changes seem reasonable given you've tested they work as before.

rickymagner (approved):

PR #113, opened 2022-09-12
QC Report updates

kachulis:
A few improvements to PRS QC report:

1. Removes control sample from batch summary table and batch score distribution plot
2. Adds table of individuals who are high in more than one condition
3. Batch score distribution plot is now interactive plotly plot
4. Batch score distribution uses data points on x axis instead of density distribution for conditions with fewer than 5 samples
5. Adds column of  number of high conditions for each individual in individual samples table, so that can sort individuals by this

### DIFF
--- a/ImputationPipeline/AggregatePRSResults.wdl
+++ b/ImputationPipeline/AggregatePRSResults.wdl
@@ -263,7 +263,7 @@ task BuildHTMLReport {
     ggplot(batch_pivoted_results %>% filter(condition %in% conditions_with_more_than_3_samples), aes(x=adjusted)) +
       geom_density(aes(color=condition), fill=NA, position = "identity") +
       xlim(-5,5) + theme_bw() + xlab("z-score") + geom_function(fun=dnorm) +
-      geom_point(data = batch_pivoted_results %>% filter(!(conditions_with_more_than_3_samples)), aes(color=condition, x = adjusted), y=0) +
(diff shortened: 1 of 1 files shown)


michaelgatzen (commented):
Some great R magic here again! Just one point, in the google doc you mentioned that the title of a report with n > 1 should be Batch ~{lab_batch} (group {n}) PRS Summary, which still needs to be changed. I'm actually wondering if it makes sense for it to contain the word "partial" somewhere, so that it's immediately clear to whoever sees that report that it's not the whole batch?

The rest looks great!

kachulis:
good call, I did forget to make that change.  I think it will be clearer as part of #114 though, so will add it there.

michaelgatzen (approved):
The code looks good, I just see that in [this submission](https://app.terra.bio/#workspaces/emerge_prod/eMerge_PRS_ck_copy/job_history/f1fac962-7bd4-48dd-9851-fe6dcfb238bb) there's still warnings being shown. I noticed that your last commit disables those warnings, maybe it's worth re-running it one more time to make sure it's working fine. Otherwise good to merge!

kachulis:
Yup, after that most recent commit the warnings are [gone](https://app.terra.bio/#workspaces/emerge_prod/eMerge_PRS_ck_copy/job_history/d836c171-96cf-4951-a310-9dd3091eb4d2).  Thanks for the review @michaelgatzen!

PR #117, opened 2022-11-21
Lowercase pcs in results csv

kachulis:
Downstream process expects PC1, PC2, etc to be pc1, pc2, etc.  Only applicable to batch results file.


michaelgatzen (approved):
Feel free to merge when the tests pass!

PR #118, opened 2023-01-27
Added documentation on Interval Files commonly used

rickymagner:
This PR introduces documentation about popular interval files used particular in various benchmarking analyses. This includes:
* adding bucket mirrors for these interval files (.bed / .interval_list),
* Terra-formatted lists to drop them into benchmarking workflows, and
* statistics relating to the overlap/redundancy between different lists on the page, along with estimates of percent of genome they cover.

Feel free to add any extra interval files to this list in similar format. Scripts for computing included statistics are also included.


Utilities/IntervalFiles/README.md
```diff
@@ -0,0 +1,237 @@
...
+* Percent of Genome: 2.7%
+* Interesting Overlaps:
+  * % of Homopolymers in LowMappability: 6% (Jaccard = 1.5%)
+  * % of Homopolymers in LowComplexityRegion: 37% (Jaccard = 26%)
+  * % of Homopolymers in SegDup: 5.6% (Jaccard = 2%)
+  * % of Homopolymers in AllDifficult: 100% (contained; Jaccard = 13%)
+
+#### LowComplexityRegion
+* Summary: Contains regions of "low complexity", using the symmetric DUST algorithm. 
+* Bucket Mirror: `gs://broad-dsde-methods-hydro-gen-truth-data-public/NIST/IntervalFiles/LCRFromHengHg38.bed`
```
meganshand, line 92: This isn't really from NIST, is it ok that you're putting it in a folder with NIST in the name?

Utilities/IntervalFiles/README.md
```diff
@@ -0,0 +1,237 @@
...
+measuring what percentage of A lies in B. This is the special case of the Jaccard index of A intersect B with A. Note also
+the percent of genome covered can be computed using a Jaccard index of your bed file A with a "full" bed file covering
+all contigs of your genome. A script to make a "full" bed from a reference dict is included in the `create_full_bed.sh`
+file. In the above calculations, a "full" bed was created from hg38 for just the autosomes and X/Y to compute what 
+percent of the genome a given interval file spanned.
+
+A script to compute all three of these (the usual Jaccard index, percent of A in B, and percent of B in A) for two 
+input bed files A and B is provided in the `overlap_stats.sh` file for reproducibility of the above code and extension 
+beyond the provided cases. You must have the `hg38-noalt.genome` or similar file for your reference in the same directory 
+when running this script for it to function properly, which is an ordered list of contigs with sequence length. Overlap
```
meganshand, line 236: Did you remove N's from the sequence length? I think it's fine either way, but it would be good to make it clear here how you got the sequence length. 

The reason I noticed this is because I think you're estimate of the percentage of the genome that the UGHiConf interval list is lower than what they report. I think this happens because they are careful to make the denominator smaller in reasonable ways. I think either way of reporting it is fine as long as it's clear.

meganshand (approved):
Thanks for writing this up @rickymagner! It will be a super helpful reference going forward.

My one concern is that as new versions of these interval lists come out it may be difficult to keep this document updated. Hopefully it will at least be straightforward to find newer versions since you provide links to the original locations and documentation.

Utilities/IntervalFiles/README.md
rickymagner: Good point, I relocated them to a better bucket path without NIST in it.

Utilities/IntervalFiles/README.md
rickymagner: Those stats were computed without removing N bases, but I added a comment to help clarify this. I did a little testing with an interval list without the Ns, and in particular, the UG one is closer to the expected 95% or so without them.

rickymagner:
I agree @meganshand about updating this with new versions. Hopefully most of them won't change too frequently (with the exception maybe of the UG one). The stats should be straightforward enough to update using the scripts included.

I just made a few quick edits now to address your helpful comments!

meganshand:
👍 Looks great, thanks @rickymagner!

PR #120, opened 2023-02-17
Added utilities for bam stats over intervals

rickymagner:
This PR includes a few additions and changes for computations over interval lists. The main components are:
* Added `ComputeIntervalBamStats.wdl`, which is a workflow that takes in a bam file, a list of interval lists, and some other metadata which computes Picard's CollectWgsMetrics on the bam over those intervals, along with a mapping quality distribution for each region. These statistics are then collected and labeled over the groups into a few output tsvs that are ready to be visualized (dashboard coming soon, maybe!). The `input_name` and `experiment` inputs are added as columns to the output, to allow concatenation of outputs across multiple sample runs and easy grouping into `experiment` groups when plotting. 
* Added a simple docker with `samtools`, which is used in the above workflow.
* Fixed some naming conventions/locations for interval files previously added. In particular, each interval file has an incarnation as a bed file, a classic interval_list with hg38 seq dict header, and an interval_list with a restricted hg38 seq dict to exclude alt-contigs. This last one was mostly relevant to some experiments I was doing without alt contigs, but might be more generally useful for workflows which try to avoid those contigs and don't want fussy tools to complain about mismatched sequence dicts.

### DIFF
--- a/Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
+++ b/Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
@@ -1,0 +1,306 @@
+version 1.0
+
+# For computing stats/distributions over various interval files for BAM files
+workflow ComputeIntervalBamStats {
+    input {
+        String input_name
+        File input_bam
+        File input_bam_index
+        String experiment = ""
+
+        File ref_fasta
+        File ref_index
+        File ref_dict
+
+        Array[File] interval_files
+        Array[String] interval_names
+    }
+
+    scatter(interval in zip(interval_names, interval_files)) {
+        call SubsetBam {
+            input:
+                input_bam=input_bam,
+                input_bam_index=input_bam_index,
+                ref_fasta=ref_fasta,
+                ref_index=ref_index,
+                ref_dict=ref_dict,
+                interval_file=interval.right,
+                interval_name=interval.left
+        }
+
+        call ComputeWgsMetrics {
+            input:
+                input_bam=SubsetBam.bam_subset,
+                input_bam_index=SubsetBam.bam_subset_index,
+                ref_fasta=ref_fasta,
+                ref_index=ref_index,
+                ref_dict=ref_dict,
+                interval_file=interval.right,
+                interval_name=interval.left
+        }
+
+        call ComputeMAPQDistribution {
+            input:
+                input_bam=SubsetBam.bam_subset,
+                input_bam_index=SubsetBam.bam_subset_index,
+                interval_file=interval.right,
+                interval_name=interval.left
+        }
+    }
+
+    call CollectCoverageResults {
+        input:
+            input_name=input_name,
+            experiment=experiment,
+            wgs_files=ComputeWgsMetrics.wgs_file
+    }
+
+    call CollectMAPQResults {
+        input:
+            input_name=input_name,
+            experiment=experiment,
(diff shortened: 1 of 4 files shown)


Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
```diff
@@ -0,0 +1,306 @@
...
+        String input_name
+        File input_bam
+        File input_bam_index
+        String experiment = ""
+
+        File ref_fasta
+        File ref_index
+        File ref_dict
+
+        Array[File] interval_files
```
tmelman, line 15: Is there a guarantee taht interval files and interval names will be in the correct order? Or is this a manual check? Or do the names not matter so much?

rickymagner: The zip in the WDL should guarantee the arrays get paired up exactly in the same order as input. Of course there's no way to prevent users from mislabeling their interval files in an order different than the one they gave (outside of using some custom structs for input, but those get a little messy to format for the input IMO).

Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
```diff
@@ -0,0 +1,306 @@
...
+        full_mapq_summary = pd.DataFrame()
+
+        for file in file_paths:
+            interval_name = os.path.basename(file).split('_mapq_dist.txt')[0]
+
+            mapq_summary = pd.read_csv(file, sep='\s+', names=['Count', 'MAPQ'])
+            mapq_summary['Interval_List'] = interval_name
+            mapq_summary['Sample'] = "~{input_name}"
+            mapq_summary['Experiment'] = "~{experiment}"
+            full_mapq_summary = pd.concat([full_mapq_summary, mapq_summary])
```
tmelman, line 233: arent you concatenating an empty dataframe (full_mapq_summary)?

Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
```diff
@@ -0,0 +1,306 @@
...
+            wgs_summary['Sample'] = "~{input_name}"
+            wgs_summary['Experiment'] = "~{experiment}"
+            full_wgs_summary = pd.concat([full_wgs_summary, wgs_summary])
+
+            cov_dist = pd.read_csv(file, sep='\t', skiprows=10)
+            cov_dist['Interval_List'] = interval_name
+            cov_dist['Sample'] = "~{input_name}"
+            cov_dist['Experiment'] = "~{experiment}"
+            full_cov_summary = pd.concat([full_cov_summary, cov_dist])
+
```
tmelman, line 288: same question as above

Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
rickymagner: For the first iteration, yes, but then the result of the concat gets assigned back to `full_mapq_summary`, so they stack over the full loop.

Utilities/IntervalFiles/ComputeIntervalBamStats.wdl
rickymagner: Same answer as above!

tmelman (approved):
LGTM

PR #122, opened 2023-03-02
AggregatePRSResults updates

kachulis:
Three improvements:

1. Adds a check that samples are only scored for allowed combinations of conditions.  Displays the number of samples scored for each allowed condition grouping, as well as listing the samples for any unallowed condition groupings.
2. In PCA plot, differentiates (by color) samples with at least 1 high risk condition from those with none.
3. Makes the threshold at which a condition is visually flagged for an unexpected control sample value a parameter of the workflow (instead of being hardcoded as 0.12)


rickymagner:
Look good to me (at least as much as I could follow the R code) and seems to do what's claimed. Also thanks for updating the tests!

rickymagner (approved):

PR #125, opened 2023-03-16
Multiqc in Terra

kachulis:


kockan (approved):
Works fine on my (cloud allocated) machine. Didn't catch anything problematic in the code. Should be mandatory fun for QC.

PR #129, opened 2023-07-06
BCFTools docker: Added the `python-is-python3` package

michaelgatzen:


rickymagner (approved):
This looks good, though I suppose if there's ever any doubt, `python3` is always safe whereas `python` might not be (if you don't have that package installed).

PR #132, opened 2023-07-31
New Utilities WDL: DownsampleAndCollectCoverage

michaelgatzen:
See Readme for description

### DIFF
--- a/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
+++ b/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
@@ -157,7 +157,7 @@ task Downsample {
 
         ~{if !defined(downsample_probability) && !defined(target_coverage) then "echo Must define either downsample_probability or target_coverage. && exit 1" else ""}
 
-        PROBABILITY=~{if defined(downsample_probability) then downsample_probability else (if select_first([target_coverage]) > select_first([original_coverage]) then "1" else "$(bc -l <<< 'scale=2; " + target_coverage + "/" + original_coverage + "')")}
+        PROBABILITY=~{if defined(downsample_probability) then downsample_probability else (if select_first([target_coverage, 0]) > select_first([original_coverage, 0]) then "1" else "$(bc -l <<< 'scale=2; " + target_coverage + "/" + original_coverage + "')")}
         
         ~{if defined(picard_jar_override) then "java -Xms2000m -Xmx2500m -jar " + picard_jar_override else 'gatk --java-options "-Xms2000m -Xmx2500m"'} \
             DownsampleSam \
--- a/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
+++ b/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
@@ -44,6 +44,8 @@ workflow DownsampleAndCollectCoverage {
             target_coverage = target_coverage,
             original_coverage = CollectOriginalCoverage.mean_coverage,
             downsample_strategy = downsample_strategy,
+            ref_fasta = ref_fasta,
+            ref_fasta_index = ref_fasta_index,
             docker = docker,
             picard_jar_override = picard_jar_override,
             preemptible = preemptible
@@ -126,6 +128,8 @@ task Downsample {
         Float? original_coverage
 
         String downsample_strategy
+        File ref_fasta
+        File ref_fasta_index
 
         String docker
         File? picard_jar_override
@@ -149,7 +153,8 @@ task Downsample {
             -O ~{output_basename}.downsampled.cram \
             -STRATEGY ~{downsample_strategy} \
             -P $PROBABILITY \
-            -CREATE_INDEX true
+            -CREATE_INDEX true \
+            -R ~{ref_fasta}
             
     }
 
--- a/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
+++ b/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
@@ -154,7 +154,7 @@ task Downsample {
             -STRATEGY ~{downsample_strategy} \
             -P $PROBABILITY \
             -CREATE_INDEX true \
-            -R ~{ref_fasta}
+            -REFERENCE_SEQUENCE ~{ref_fasta}
             
     }
 
--- a/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
+++ b/Utilities/WDLs/DownsampleAndCollectCoverage.wdl
@@ -165,6 +165,10 @@ task Downsample {
 
         ~{if !defined(downsample_probability) && !defined(target_coverage) then "echo -e '\nERROR: Must define either downsample_probability or target_coverage.\n'; exit 1" else ""}
 
+        # If downsample_probability defined then use that, otherwise use min(1, target_coverage / original coverage).
+        # Note that we have to use select_first with a default of zero to be able to evaluate the input validation above.
+        # If we wouldn't default to zero then the script evaluation would fail in Cromwell already and the error message
+        # above would not be printed.
         PROBABILITY=~{if defined(downsample_probability) then downsample_probability else (if select_first([target_coverage, 0]) > select_first([original_coverage, 0]) then "1" else "$(bc -l <<< 'scale=2; " + target_coverage + "/" + original_coverage + "')")}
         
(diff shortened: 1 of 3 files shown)


michaelgatzen:
@rickymagner I checked the binary file header of the output file and it's a valid CRAM file

rickymagner:
Yeah I checked that it was `MergeSamFiles` that didn't write crams properly

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,172 @@
...
+        CollectWgsMetrics \
+        -INPUT ~{input_cram} \
+        -VALIDATION_STRINGENCY SILENT \
+        -REFERENCE_SEQUENCE ~{ref_fasta} \
+        ~{"-INTERVALS " + coverage_intervals} \
+        -OUTPUT ~{output_basename}.wgs_metrics \
+        -READ_LENGTH ~{read_length} \
+        -USE_FAST_ALGORITHM ~{use_fast_algorithm}
+
+        cat ~{output_basename}.wgs_metrics | grep -v -e "^#" -e "^$" -e "^GENOME_TERRITORY" | head -n 1 | awk '{ print $2 }' > ~{output_basename}.mean_coverage
```
rickymagner, line 109: I can see this works though seeing the constant string `GENOME_TERRITORY` feels a bit messy. I don't think it's necessary. And the `$2` is technically not robust against changes in Picard's output format (as if that would ever happen), but at least this should still work:

```
grep -v -e '^#' -e "^$" "~{output_basename}.wgs_metrics" | head -2 | awk '{ print $2 }' | tail -1
```
If you really wanted to avoid the hardcoded `$2` in the awk script, you could do something like:
```
grep -v -e '^#' -e "^$" "~{output_basename}.wgs_metrics" | head -2 > wgs.tsv
COL_NUM=$(sed 's/\t/\n/g' | grep -n 'MEAN_COVERAGE' | cut -d':' -f1)
awk -v col=$COL_NUM ' { print $col }' wgs.tsv | tail -1 > "~{output_basename}.mean_coverage"
```
but I think this is probably more trouble than it's worth.

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,172 @@
...
+            -STRATEGY ~{downsample_strategy} \
+            -P $PROBABILITY \
+            -CREATE_INDEX false \
+            -REFERENCE_SEQUENCE ~{ref_fasta}
+        
+        samtools index ~{output_basename}.downsampled.cram
+    }
+
+    runtime {
+        docker: docker # TODO: update docker to use the new Picard options
```
rickymagner, line 162: Was there a TODO missed here?

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,172 @@
...
+    }
+    Int disk_size = ceil(3.5 * size(input_cram, "GiB") + 20 + additional_disk_gb)
+
+    String output_basename = sub(sub(basename(input_cram), "\\.bam$", ""), "\\.cram$", "")
+
+
+    command {
+        set -e -o pipefail
+
+        PROBABILITY=~{if defined(downsample_probability) then downsample_probability else "$(bc -l <<< 'scale=2; " + target_coverage + "/" + original_coverage + "')"}
```
rickymagner, line 147: You might want to handle the edge case if `target_coverage` is accidentally larger than `original_coverage` to set this to 1.

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,172 @@
...
+        ~{if defined(picard_jar_override) then "java -Xms2000m -Xmx2500m -jar " + picard_jar_override else 'gatk --java-options "-Xms2000m -Xmx2500m"'} \
+            DownsampleSam \
+            -I ~{input_cram} \
+            -O ~{output_basename}.downsampled.cram \
+            -STRATEGY ~{downsample_strategy} \
+            -P $PROBABILITY \
+            -CREATE_INDEX false \
+            -REFERENCE_SEQUENCE ~{ref_fasta}
+        
+        samtools index ~{output_basename}.downsampled.cram
```
rickymagner, line 158: Why not use `-CREATE_INDEX true` ?

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,172 @@
...
+            use_fast_algorithm = use_fast_algorithm,
+            docker = docker,
+            picard_jar_override = picard_jar_override,
+            preemptible = preemptible
+    }
+
+    output {
+        File downsampled_cram = Downsample.downsampled_cram
+        File downsampled_cram_index = Downsample.downsampled_cram_index
+        Float downsampled_mean_coverage = CollectDownsampledCoverage.mean_coverage
```
rickymagner, line 71: Do you think it makes sense to also return the `wgs_metrics` files used to compute the mean coverage? Sometimes those other stats (like the variance of coverage, etc) are also useful to have in a Terra table.

rickymagner (commented):
I'm going to definitely use this WDL a bunch once it's done, so thanks for creating it! I added a few comments and suggestions here.

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -104,18 +104,25 @@ task CollectWgsMetrics {
         ~{"-INTERVALS " + coverage_intervals} \
         -OUTPUT ~{output_basename}.wgs_metrics \
         -READ_LENGTH ~{read_length} \
-        -USE_FAST_ALGORITHM ~{use_fast_algorithm}
+        -USE_FAST_ALGORITHM ~{use_fast_algorithm} \
+        -COUNT_UNPAIRED true
+
+        cat ~{output_basename}.wgs_metrics | grep -v -e '^#' -e "^$" "~{output_basename}.wgs_metrics" | head -2 > wgs.tsv
```
rickymagner, line 113: This line shouldn't have the `cat` statement since `grep` is reading directly from the file

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -104,18 +104,25 @@ task CollectWgsMetrics {
...
+        COL_NUM=$(sed 's/\t/\n/g' | grep -n 'MEAN_COVERAGE' | cut -d':' -f1)
```
rickymagner, line 114: In this case, the `sed` command should have the `wgs_metrics` file as its last argument. This is basically just splitting all the tab-delimited sections into newlines, and then asking `grep` which line number contains `MEAN_COVERAGE` in order to get the column number it's one, so that `awk` can just print that one.

rickymagner: Also I think this is still overkill but just saw it as a proof of concept, in case you have a simpler way to actually do it :)

rickymagner (commented):
Just a few more comments!

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -0,0 +1,194 @@
...
+    runtime {
+        docker: docker
+        preemptible: preemptible
+        memory: "14 GiB"
+        cpu: "16"
+        disks: "local-disk " + disk_size + " HDD"
+    }
+
+    output {
+        File downsampled_cram = output_basename + ".downsampled." +output_extension
```
kockan, line 191: Whitespace:
`File downsampled_cram = output_basename + ".downsampled." + output_extension`

kockan (approved):
LGTM.

rickymagner (approved):
I approve (after that comment above is added)!

Utilities/WDLs/DownsampleAndCollectCoverage.wdl
```diff
@@ -181,6 +181,11 @@ task Downsample {
             -CREATE_INDEX false \
             -REFERENCE_SEQUENCE ~{ref_fasta}
         
+        # Create reference cache for indexing CRAM file
+        seq_cache_populate.pl -root ./ref/cache ~{ref_fasta}
```
rickymagner, line 185: Wait where does this script come from? Is it already in the docker?

michaelgatzen: @rickymagner Yep, `apt install samtools` installs it into `/usr/bin`

PR #133, opened 2023-08-03
Handle ploidy in chrX and chrY outside of the PAR region

ty-kay:
…egion


Utilities/Dockers/VCFEval/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN mv bedtools.static bedtools && chmod +x bedtools && cp bedtools /bin/
+
+# Make bcftools
+ADD https://github.com/samtools/bcftools/releases/download/1.16/bcftools-1.16.tar.bz2 .
+RUN tar -xvf bcftools-1.16.tar.bz2
+WORKDIR bcftools-1.16
+RUN ./configure
+RUN make
+RUN make install
+RUN export PATH=/bcftools-1.16/bcftools:$PATH
```
michaelgatzen, line 33: `make install` does this
```suggestion
```

michaelgatzen (changes requested):

.gitignore
```diff
@@ -32,8 +32,8 @@ old_*
 *.bai
 *.idx
 *.vcf
-*.vcf.gz
-*.tbi
+#*.vcf.gz
+#*.tbi
```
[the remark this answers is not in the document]
rickymagner: Yeah given the other comment this makes sense. I reverted it

BenchmarkVCFs/PARTest/README.md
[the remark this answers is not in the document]
rickymagner: That's a good point. I think I originally had Ty put the stuff all here since he didn't have Github access for a while to our private repo. I'll relocate this stuff over to an analysis there.

Utilities/Dockers/VCFEval/Dockerfile
rickymagner: Nice

BenchmarkVCFs/SimpleBenchmark.wdl
```diff
@@ -264,52 +267,126 @@ task VCFEval {
...
+                -o reg
+
+            mkdir output_dir
+            for file_name in "tp-baseline" "tp" "fp" "fn";
+            do
+                bcftools merge --force-samples "par/${file_name}.vcf.gz" "reg/${file_name}.vcf.gz" | bcftools sort -Oz -o "output_dir/${file_name}.vcf.gz"
+                bcftools index -t "output_dir/${file_name}.vcf.gz"
+            done
+
+        fi
```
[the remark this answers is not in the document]
rickymagner: The first part of the script is a data cleaning step, so it should run in both cases. Then the second half handles the combining stats logic if a PAR file is provided.

BenchmarkVCFs/SimpleBenchmark.wdl
```diff
@@ -264,52 +267,126 @@ task VCFEval {
...
+                -t rtg_ref \
+                -o reg \
+
+            mkdir output_dir
+            cp reg/*.vcf.gz* output_dir/
+
+        else
+            # Handle case where user provides PAR bed by running with --squash-ploidy over haploid region
+            grep -E '^@SQ' ~{reference.dict} | awk 'BEGIN{OFS="\t"} {gsub(/SN:/,"",$2); gsub(/LN:/,"",$3); print $2, $3}' > 'genome_file.txt'
+            bedtools complement -i ~{par_bed} -g genome_file.txt -L > 'par.bed'
```
kachulis, line 294: this ended up confusing me a bit, I think because of the name of the output.  If you name the output here `non-par.bed` instead of `par.bed` I think it would be clearer, since the output is the non-par region, if I'm understanding correctly.

BenchmarkVCFs/SimpleBenchmark.wdl
```diff
@@ -264,52 +267,126 @@ task VCFEval {
...
+                -b ~{base_vcf} \
+                -c ~{call_vcf} \
+                --bed-regions par.bed \
+                ~{"-e " + evaluation_bed} \
+                --vcf-score-field="~{score_field}" \
+                --output-mode split \
+                --decompose \
+                --roc-subset snp,mnp,indel \
+                -t rtg_ref \
+                -o par
```
kachulis, line 312: similarly, would name this `non-par` instead (unless I'm just completely turned around right now, which is possible!)

Utilities/Dockers/VCFEval/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:22.10
```
kachulis, line 1: it there a need to keep [the rtg dockerfile](https://github.com/broadinstitute/palantir-workflows/blob/main/Utilities/Dockers/RTG/Dockerfile) around, or can this replace it?

kachulis (commented):
just a couple small comments, otherwise looks great!

BenchmarkVCFs/SimpleBenchmark.wdl
rickymagner: Yes, I think you're right, and updated the name accordingly.

BenchmarkVCFs/SimpleBenchmark.wdl
rickymagner: Yes!

Utilities/Dockers/VCFEval/Dockerfile
rickymagner: I think it depends on how we want to organize the dockers in the repo overall. Originally, it started from the philosophy of trying to have a bunch of lightweight docker images with "just" what you need for various tasks. But since often WDLs end up with tasks that combine lots of tools together, it might make sense to also add some of these beefier images to our repertoire as well, or even as replacement. 

I don't have strong opinions either way, other than I have been considering making a large "all-in-one" image for people who don't want to micromanage their environments, in which case this could be included in.

kachulis (approved):
Looks good, thanks @ty-kay and @rickymagner!

rickymagner:
It looks like merging is blocked until @michaelgatzen is back and can approve the changes he requested

michaelgatzen (approved):

PR #137, opened 2023-08-10
Update for PRS on hg38

kachulis:
A few improvements, and enabling scoring on hg38

1. plink needs to be explicitly told to allow `chr` on the beginning of chromosome names.  The pipeline now looks at the weights file to automatically determine the chromosome naming pattern that should be passed to plink.
2. an option is now available to code sites as `chr:pos:ref:alt` instead of `chr:pos:a1:a2` (where a1, a2 are in lexicographical order).  This allows indels to be included which could otherwise be ambiguous (if a vcf includes an A--> AA insertion and an AA --> A deletion at the same site, for example).  The default is still to use `chr:pos:a1:a2`, not because this is the encouraged method (it isn't and should be avoided if possible), but to avoid changing default behavior, at least for the time being.
3. there was a bug which meant that a list of sites to subset to before scoring was effectively required, even though it wasn't supposed to be.  this is fixed now.

### DIFF
--- a/ImputationPipeline/ScoringTasks.wdl
+++ b/ImputationPipeline/ScoringTasks.wdl
@@ -648,3 +648,36 @@ task ExtractIDsPlink {
     memory: mem + " GB"
   }
 }
+
+#plink chromosome encoding rules: https://www.cog-genomics.org/plink/2.0/data#irreg_output
+task DetermineChromosomeEncoding {
+  input {
+    File weights
+  }
+
+  command <<<
+    python3 << "EOF"
+      with open(~{weights}) as weights_file:
+      chroms = {s.split("\t")[0].split(":")[0] for s in weights_file if ":" in s}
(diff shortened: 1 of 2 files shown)


ImputationPipeline/ScoringTasks.wdl
```diff
@@ -645,3 +648,36 @@ task ExtractIDsPlink {
...
+task DetermineChromosomeEncoding {
+  input {
+    File weights
+  }
+
+  command <<<
+    python3 << "EOF"
+    code = 'MT'
+    with open("~{weights}") as weights_file:
+      chroms = {s.split("\t")[0].split(":")[0] for i, s in enumerate(weights_file) if i > 0}
```
michaelgatzen, line 662: This is a really cool trick

michaelgatzen (approved):

Issue #67, opened 2021-09-14
Add an error message/potentially rethink this error check


tmelman:
https://github.com/broadinstitute/palantir-workflows/blob/d750284988b616894fc28181e40e9075a0131ecf/ImputationPipeline/PRSWrapper.wdl#L105

At this line, if the VCF name is not the same as the filename, the script fails silently. Address this by adding an error message on failure, and make sure this is the intended behavior.

kachulis:
Could also refactor to read sample_id from vcf instead of being passed as input.

rickymagner:
Oh, if you put in a multi-sample VCF maybe you want the user to be able to point to which sample to use? Is that ever a reasonable use case for this workflow?

In any case, I just made a quick fix to print error messages rather than quit with no helpful message.

rickymagner:
I'm going to close for now given #141 since it might be useful to still allow the user to provide a `sample_id` when using a multi-sample VCF. Feel free to reopen if you think it might be better to add some logic to infer the name from the VCF itself (e.g. default to the first sample, or something else)

PR #141, opened 2023-08-25
Add error messages in R scripts rather than just quit for PRSWrapper

rickymagner:
Addresses #67. 

Note that while it seems the `stop` function in R doesn't allow for custom error codes like `quit`, the default exit code for `stop` is still 1, so nothing has changed except the inclusion of some more helpful error messages.


meganshand (approved):
👍

Issue #138, opened 2023-08-25
SimpleBenchmark doesn't run without providing evaluation bed file


rickymagner:
The `evaluation_bed` input used for `vcfeval` in `SimpleBenchmark` does not work properly. It is supposed to be an optional file input, but the WDL always crashes when running in Terra if you don't provide it.

PR #142, opened 2023-08-28
Update RTG docker for vcfeval to fix FORMAT filters ignored

rickymagner:
This short PR bumps the rtg/vcfeval version used in BenchmarkVCFs so that it uses rtg 3.12.1. This version added support for vcfeval to be aware of sample-level filters (FORMAT/FT) when making comparisons. Previously, if a *site* was PASS but a sample had a non-passing FT field, it would still be counted towards the comparison. 

Also, cleaned up the Python in that block of code slightly. :)

### DIFF
--- a/BenchmarkVCFs/BenchmarkVCFs.wdl
+++ b/BenchmarkVCFs/BenchmarkVCFs.wdl
@@ -546,8 +546,8 @@ task VcfEval {
     command <<<
     set -xeuo pipefail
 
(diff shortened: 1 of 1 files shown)


rickymagner:
This also addresses my half of [this issue](https://github.com/broadinstitute/palantir-workflows/issues/138#issuecomment-1693617360) and makes `SimpleBenchmark` and `BenchmarkVCFs` in sync scientifically (more or less).

PR #144, opened 2023-08-31
Dd toggle benchmark vc fs fingerprint

ddsva1:
Added a variable to turn off the picard fingerprinting when running BenchmarkVCFs.wdl; please let me know if there are any issues or improvements to make!
Essentially, if the boolean 'fingerprint_toggle' is false, then the MatchEvalTruth task is skipped, but if the variable is true or blank then BenchmarkVCFs runs as usual.
Thank you!
-Dill :]

### DIFF
--- a/BenchmarkVCFs/BenchmarkVCFs.wdl
+++ b/BenchmarkVCFs/BenchmarkVCFs.wdl
@@ -62,7 +62,7 @@ workflow Benchmark {
         vcfScoreField: {description:"Have vcfEval use this field for making the roc-plot. If this is an info field (like VSQLOD) it should be provided as INFO.VQSLOD, otherewise it is assumed to be a format field."}
         gatkJarForAnnotation: {description:"GATK jar that can calculate necessary annotations for jexl Selections when using VCFEval."}
         annotationNames: {description:"Annotation arguments to GATK (-A argument, multiple OK)"}
-        fingerprint_toggle: {description:"no input or true toggles on fingerprinting, false toggles off fingerprinting"}
+        fingerprint_toggle: {description:"No-input or true keeps fingerprinting on, but false turns fingerprinting off."}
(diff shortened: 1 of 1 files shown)


BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -31,6 +31,7 @@ workflow Benchmark {
         Boolean passingOnly=true
         String? vcfScoreField
         String? dummyInputForTerraCallCaching
+        Boolean fingerprint_toggle = true
```
rickymagner, line 34: I would name this something where it's clear what true vs false means, e.g. `run_fingerprint = true` means to run the fingerprint task.

BenchmarkVCFs/BenchmarkVCFs.wdl
```diff
@@ -96,16 +98,17 @@ workflow Benchmark {
...
-    call MatchEvalTruth as Match {
-        input:
-            evalVcf=evalVcf,
-            truthVcf=truthVcf,
-            evalVcfIndex=evalVcfIndex,
-            truthVcfIndex=truthVcfIndex,
-            hapMap=hapMap,
-            gatkTag=gatkTag,
-            preemptible=preemptible
+    if (fingerprint_toggle){
```
rickymagner, line 101: I'd put a space between the `)` and the `{` so `(run_fingerprint) {` instead

rickymagner (commented):
Looks good with minor comments! I'm assuming you tested it works / no other tasks are depending on the output for that task?

BenchmarkVCFs/BenchmarkVCFs.wdl
ddsva1: *{,  not }, for the commit message below!

BenchmarkVCFs/BenchmarkVCFs.wdl
ddsva1: Changed it to runFingerprintCheck - remembered the camel case!

rickymagner: For variables in WDL it's also "ok" to use snake_case, but this works too. The last time I mentioned using CapitalCamelCase for tasks/workflow names since those are "big" object blocks.

rickymagner (approved):
Looks good to me!

ddsva1:
"Looks good with minor comments! I'm assuming you tested it works / no other tasks are depending on the output for that task?"
There isn't an output for that task in the wdl, and the workflow runs fine when it is removed (it just skips the Match task when runFingerprintCheck is false, but runs that task when true or blank)

PR #154, opened 2023-09-26
Add utility for converting interval_list to bed efficiently

rickymagner:
This short PR adds a utility that can convert a mix of `.interval_list` and `.bed` files into a list of just `.bed` files, which is ideal for pipelines using tools that only take in `.bed` files. The WDL is designed so that if all the inputs are `.bed` files, then no tasks are called (just some simple WDL code is run), and the original list and labels are returned. This means pipelines accepting just `.bed` files can use this task as a penalty-free (for `.bed` file users) way to extend their functionality to `.interval_list`s as well. The task also allows for optional label inputs so that labels can stay "next to" their converted file and used downstream. 

See the README in this PR for more details. This utility will be used in the upcoming benchmarking WDL and others.

### DIFF
--- a/Utilities/WDLs/Interval2Bed.wdl
+++ b/Utilities/WDLs/Interval2Bed.wdl
@@ -59,21 +59,19 @@ workflow Interval2Bed {
 task Convert {
     input {
         File interval_list
-        String? interval_label
+        String interval_label
 
-        String gatk_tag = "4.2.3.0"
+        String gatk_tag = "4.4.0.0"
         Int? preemptible
         Int disk_size = ceil(4 * size(interval_list, "GB")) + 5
         Int cpu = 4
         Int memory = 16
     }
 
-    String name = basename(interval_list, ".interval_list")
-
     command <<<
         set -xe
 
-        gatk IntervalListToBed -I ~{interval_list} -O ~{name}.bed
+        gatk IntervalListToBed -I ~{interval_list} -O ~{interval_label}.bed
     >>>
 
     runtime {
@@ -86,7 +84,7 @@ task Convert {
     }
 
     output {
-        File bed_file = "~{name}.bed"
-        String? bed_label = select_first([interval_label, name])
+        File bed_file = "~{interval_label}.bed"
+        String bed_label = interval_label
     }
 }
--- a/Utilities/WDLs/Interval2Bed.wdl
+++ b/Utilities/WDLs/Interval2Bed.wdl
@@ -1,0 +1,92 @@
+version 1.0
+
+workflow Interval2Bed {
+    input {
+        Array[File] interval_files
+        Array[String]? interval_labels
+    }
+
+    if (defined(interval_labels)) {
+        Array[Pair[String, File]] paired_intervals = zip(select_first([interval_labels]), interval_files)
+
+        scatter (p in paired_intervals) {
+            # Check that file extension is .interval_list, i.e. removing it shortens the basename
+            if (basename(p.right, ".interval_list") != basename(p.right)) {
+                call Convert as ConvertPairs {
+                    input:
+                        interval_list=p.right,
+                        interval_label=p.left
+                }
+            }
+
+            if (basename(p.right, ".bed") != basename(p.right)) {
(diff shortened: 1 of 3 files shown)


Utilities/WDLs/Interval2Bed.wdl
```diff
@@ -0,0 +1,92 @@
...
+        Array[File] interval_files
+        Array[String]? interval_labels
+    }
+
+    if (defined(interval_labels)) {
+        Array[Pair[String, File]] paired_intervals = zip(select_first([interval_labels]), interval_files)
+
+        scatter (p in paired_intervals) {
+            # Check that file extension is .interval_list, i.e. removing it shortens the basename
+            if (basename(p.right, ".interval_list") != basename(p.right)) {
```
meganshand, line 14: nice way to make this check without spinning up a task 👍

Utilities/WDLs/Interval2Bed.wdl
```diff
@@ -0,0 +1,92 @@
...
+                File bed_file_pair = p.right
+                File bed_label_pair = p.left
+            }
+        }
+
+        Array[File] new_interval_files_pair = flatten([select_all(bed_file_pair), select_all(ConvertPairs.bed_file)])
+        Array[File] new_interval_labels_pair = flatten([select_all(bed_label_pair), select_all(ConvertPairs.bed_label)])
+    }
+
+    if (!defined(interval_labels)) {
```
meganshand, line 32: I was trying to think through if there was a way to have one block without this `defined` check using a bunch of `select_first(p.left, basename(p.left, "interval_list)` but having the zip makes handling that a bit ugly right? I'm assuming you also thought through that option and concluded that this is cleanest way to do this?

Utilities/WDLs/Interval2Bed.wdl
```diff
@@ -0,0 +1,92 @@
...
+        Array[String] bed_labels = select_first([new_interval_labels_pair, new_interval_labels])
+    }
+}
+
+task Convert {
+    input {
+        File interval_list
+        String? interval_label
+
+        String gatk_tag = "4.2.3.0"
```
meganshand, line 64: Is there a reason not to use 4.4.0.0 as the default?

Utilities/WDLs/Interval2Bed.wdl
```diff
@@ -0,0 +1,92 @@
...
+    output {
+        Array[File] bed_files = select_first([new_interval_files_pair, new_interval_files])
+        Array[String] bed_labels = select_first([new_interval_labels_pair, new_interval_labels])
+    }
+}
+
+task Convert {
+    input {
+        File interval_list
+        String? interval_label
```
meganshand, line 62: Aren't you always providing this in the WDL so it could just be `String` instead of optional? Meaning the output `bed_label` could also be not optional?

meganshand (approved):
This looks great! Thanks for the excellent documentation as well. Couple of small questions.

Utilities/WDLs/Interval2Bed.wdl
rickymagner: Yeah I'm not sure if there are some WDL tricks to make this a bit more elegant, but thought this might be the simplest. I really needed to be able to keep the labels next to the files to ensure they get paired in the outputs as well, and this felt like two different cases. If you have some ideas I'd definitely like to make this nicer!

Utilities/WDLs/Interval2Bed.wdl
rickymagner: Good catch! That's a leftover from old code.

Utilities/WDLs/Interval2Bed.wdl
rickymagner: That's a good point. Also updated to change this.

Utilities/WDLs/Interval2Bed.wdl
meganshand: sounds good, leave it as is then. I think all of my ideas just ultimately made things messier.

PR #156, opened 2023-11-07
Add MatchFingerprints utility WDL

rickymagner:
This PR adds a utility for handling various applications of fingerprinting. In particular, it should have functionality to perform the following:
* Act as a failsafe, allowing you to fail workflows when paired files don't match fingerprinting as expected;
* Allow you to match across all pairs in two lists of files, for when you might not be particularly sure of how to match up files due to known/suspected sample swaps, easy benchmarking truth-matching, or other cases;
* Give users the ability to toggle a few specific details, like forcing sample names to match in files, toggling the threshold for fingerprinting matches, etc.

It should be simple to drop into WDLs which require matched files for the same samples (like benchmarking query vs truth data), and use the resulting `matched_pairs` output of the task to ensure you only act on fingerprint-matched pairs of files. See the README edits for some more details.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ executors:
   womtool-executor:
     docker:
       # specify the version you desire here
-      - image: broadinstitute/womtool:71
+      - image: broadinstitute/womtool:84
         entrypoint: /bin/bash
     working_directory: ~/palantir-workflows
     environment: WOMTOOL_JAR=/app/womtool.jar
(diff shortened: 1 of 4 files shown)


rickymagner:
Just made some changes to try to simplify the code and also require indices. I tried making index files optional, but this led to the discovery of a Cromwell bug for one particular implementation (issue opened in their repo), so I abandoned that for now.

There has been some ongoing discussion about a possible Terra bug where GATK cannot stream from a requester-pays bucket. This is particularly important for this workflow in being used for fingerprinting samples against our NIST requester-pays mirror, so I'll try to follow up on that in the future depending on how the Terra support ticket resolves, which may or may not require changes to this code to resolve. For now, this workflow should work fine streaming any files from normal GCP buckets.

kockan (changes requested):
Looks good to me. Ran a test on a Terra workspace and it worked beautifully. Couldn't find it on Dockstore though and had to copy/paste the WDL to my methods repo so I believe dockstore.yml needs to be updated before merging.

rickymagner:
Good catch. Just updated the dockstore yml now

kockan (approved):

PR #161, opened 2023-12-13
Add Checkpointing and QC Collection to GlimpseImputation

michaelgatzen:
- Add checkpointing and resource selection to Glimpse2Imputation
- Add task CollectQCMetrics to Glimspe2Imputation to generate metrics TSV output
- Add merging of QC metrics to Glimpse2MergeBatches
- Add safety check of colliding CRAM filenames to Glimpse2Imputation
- Add new workflow: Glimpse2CheckQC
- Add shell script to build docker image extension from GLIMPSE fork and branch
- Remove default arguments in Glimpse2ImputationInBatches so that the default values of the underlying workflows propagate up
- Add documentation


michaelgatzen:
TODO: Documentation on how to build docker images

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -5,42 +5,68 @@ workflow Glimpse2MergeBatches {
         Array[File] imputed_vcfs
         Array[File] imputed_vcf_indices
 
+        Array[File?]? qc_metrics
```
kachulis, line 8: is it really necessary for this to be `Array[File?]?` as opposed to just `Array[File]?` 

I get that it's slightly more flexible, but the use case of having qc_metrics for some but not all of the batches doesn't seem particularly useful.  And honestly, would we want to allow that in the first place?  or does it make more sense to say that if your passing in metrics files to merge, you should have them for every batch?

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
```diff
@@ -0,0 +1,127 @@
...
+fi
+
+trap cleanup SIGINT SIGTERM ERR EXIT
+
+if docker images | grep "temp_glimpse_base" > /dev/null; then
+    die "temp_glimpse_base Docker image exists already. Please remove it before building to ensure that the correct base image is used: docker image rm temp_glimpse_base"
+    exit 1
+fi
+
+git clone $repo --branch $branch --single-branch glimpse_base
```
kachulis, line 119: does this need to be `${script_dir}/glimpse_base`

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env bash
+
+# This branch is using the following configuration.
+# To replicate this branch, run this script with the following arguments:
+# build_base_and_extension_docker.sh -r "https://github.com/kachulis/GLIMPSE.git" -b ck_checkpoint_clean -t "glimpse_build_test"
+
+set -Eeuo pipefail
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
```
kachulis, line 9: why would `script_dir=$(dirname "${BASH_SOURCE[0]}")` not work?

kachulis (commented):
a few initial comments, but generally looks great!

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
michaelgatzen: To be completely honest, I used a bash script template for this, but I did some research and according to [this](https://www.ostricher.com/2014/10/the-right-way-to-get-the-directory-of-a-bash-script/), using just `BASH_SOURCE` would break if there are symlinks involved

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
michaelgatzen: Yes, you're right. Fixed

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
michaelgatzen: Yes, I agree that metrics should not be passed for some and not all batches. The reason for the `File?` is that if `collect_qc_metrics` is set to false then Glimpse2Imputation returns null as the `qc_metrics` output, which means that Glimpse2MergeBatches actually receives `[null, null]` in the case of two batches. That's why I implemented it this way to keep it as flexible as possible, but I'm debating to implement it differently, by making it an `Array[File]` and calling it either with or without that argument from Glimpse2ImputationInBatches

GlimpseImputationPipeline/Glimpse2ImputationInBatches.wdl
```diff
@@ -70,26 +70,45 @@ workflow Glimpse2ImputationInBatches {
...
+    if (!merge_qc_metrics) {
+        call MergeBatches.Glimpse2MergeBatches as MergeBatchesWithoutQCMetrics {
+            input:
+                imputed_vcfs = Glimpse2Imputation.imputed_vcf,
+                imputed_vcf_indices = Glimpse2Imputation.imputed_vcf_index,
+                output_basename = output_basename,
+                docker_extract_annotations = docker_extract_annotations,
+                docker_count_samples = docker_count_samples,
+                docker_merge = docker_merge
+        }
```
kachulis, line 100: I think you could also do this this way, which feels maybe slightly clearer?

```suggestion
    if (merge_qc_metrics) {
        Array[File] qc_metrics = select_all(Glimpse2Imputation.qc_metrics)
    }
    call MergeBatches.Glimpse2MergeBatches as MergeBatchesWithQCMetrics {
        input:
            imputed_vcfs = Glimpse2Imputation.imputed_vcf,
            imputed_vcf_indices = Glimpse2Imputation.imputed_vcf_index,
            output_basename = output_basename,
            qc_metrics = qc_metrics,
            docker_extract_annotations = docker_extract_annotations,
            docker_count_samples = docker_count_samples,
            docker_merge = docker_merge
    }
```

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -52,20 +52,19 @@ workflow Glimpse2MergeBatches {
 
     # This handles the qc_metrics in the case of only one batch. If qc_metrics is not
     # defined then this if statement will be false and qc_metrics_1 will be null.
-    # If qc_metrics is defined then set qc_metrics_1 to the first element, which
-    # may or may not be null.
+    # If qc_metrics is defined then set qc_metrics_1 to the first element
     if (length(imputed_vcfs) == 1 && defined(qc_metrics)) {
-        File? qc_metrics_1 = select_first([qc_metrics, []])[0]
+        File qc_metrics_1 = select_first([qc_metrics, []])[0]
```
kachulis, line 57: since you're requiring qc_metrics be defined here, you don't need the extra empty array, do you?  ie, `select_first([qc_metrics])[0]` should work, right?

kachulis (approved):
I know this is still marked as draft, but it all looks good to me, just a couple minor additional suggestions.  If you make substantial additional changes I (or someone else) can re-review.

PR #163, opened 2024-03-18
Update IntervalList2Bed to handle case of compressed interval file inputs

rickymagner:
Updates the utility to convert `.interval_list` to `.bed` files to allow for `.gz` compressed input files. Currently the workflow will ignore / remove any files that end in `.gz`.


kockan (approved):
Looks good to go

PR #164, opened 2024-03-22
Implement FindSamplesAndBenchmark wrapper for SimpleBenchmark

rickymagner:
This PR adds `FindSamplesAndSimpleBenchmark` in staging in anticipation of eventually replacing `FindSamplesAndBenchmark`. The WDL accepts a list of base_vcfs and query_vcfs, performs fingerprinting across all pairs (default; can also just compare across equal index in lists), and then runs `SimpleBenchmark` on the matching samples. It also allows for cohorts to be compared, and matching samples within each pair leads to another run of `SimpleBenchmark` in the current implementation.

A few auxiliary utilities are added/updated to help with this:
- CombineTables: takes an array of tsv's and just concats them using pandas
- MatchFingerprints: updated to carry some metadata associated with the input files to keep evaluation_bed files "next to" their corresponding truth files. Something like this is needed since the original arrays will be randomly subset based on the results of fingerprinting, but it's a bit of a hack. When WDL 1.1 introduces better support for making `Map[K, V]` types this could be better handled.

There are a couple of WDL hacks to try to keep things as streamlined as possible without spinning up extra tasks, but that does hurt readability in a few places. If anyone has any suggestions for better implementations without introducing new tasks, I'd appreciate it. Otherwise it could be revisited once Terra supports WDL 1.1.

The workflow was tested on matching samples from a cohort to a list of GIAB truth data.


.dockstore.yml
```diff
@@ -111,3 +111,6 @@ workflows:
   - name: DetectPCANovelties
     subclass: WDL
     primaryDescriptorPath: /Utilities/WDLs/DetectPCANovelties.wdl
+  - name: FindSamplesAndSimpleBenchmark
```
meganshand, line 114: Should we remove `FindSamplesAndBenchmark` from dockstore at this point? Or are there still features one might choose to use the old version for?

BenchmarkVCFs/dockstore_sync.txt
```diff
@@ -0,0 +1 @@
+test
```
meganshand, line 1: Did you mean to commit this?

BenchmarkVCFs/FindSamplesAndSimpleBenchmark.wdl
```diff
@@ -0,0 +1,146 @@
...
+        File ref_fasta
+        File ref_index
+
+        # Subsetting inputs using intervals
+        Array[File] stratifier_intervals = []
+        Array[String] stratifier_labels = []
+
+        # Evaluation inputs
+        Array[File] evaluation_intervals
+        String score_field = "GQ"
```
meganshand, line 38: Is score field usually in the FORMAT field? I'd have guessed that the default here should have been QUAL or VQSLOD or some other INFO level field?

.dockstore.yml
rickymagner: The plan is to eventually replace it completely with this WDL. It would have to be done at the same time as replacing BenchmarkVCFs => SimpleBenchmark code, so this is more a "staging" PR and that merge will happen after checking with enough people that breaking main won't ruin all their dockstore imports (they should pin the version).

BenchmarkVCFs/dockstore_sync.txt
rickymagner: Good catch; removed.

BenchmarkVCFs/FindSamplesAndSimpleBenchmark.wdl
rickymagner: I kept the default to GQ since that's the default for the flag in vcfeval itself

meganshand (approved):
👍

Issue #80, opened 2021-11-04
FunctionalEquivalence cannot handle data with no QUAL < 30 calls made


rickymagner:
This bug was discovered using possibly dubious data, but it seems that if one tries to use FE on vcfs where QUAL >= 30 everywhere, then it won't yield the appropriate F1 plots at the end (they're just empty since no data is collected to compare for such high quality calls). It won't even produce a default "line" graph as output, which might be worth putting in. I'm not sure this is a major issue if one expects to always have at least a good portion of QUALs < 30. Maybe make this threshold an input with default value 30?

rickymagner:
Resolved by #167

Issue #49, opened 2021-07-26
FunctionalEquivalence: Passing information through file names


michaelgatzen:
Some metadata is passed through the workflow by using filenames, which prohibits certain characters (e.g. a period in the dataset column). Solve this by passing this information in a different way.

rickymagner:
Resolved by #167

PR #110, opened 2022-08-22
Revert "FE: rewrote F1 evaluation using Pandas"

michaelgatzen:
Reverts broadinstitute/palantir-workflows#97

PR #167, opened 2024-04-10
The Big Benchmarking PR

rickymagner:
This PR completes the thread of revamping our benchmarking suite of tools started sometime ago. In particular, it replaces `BenchmarkVCFs` with the `SimpleBenchmark` rewrite and updates its dependencies accordingly. Documentation is updated, and a few other utilities are shuffled around and refactored to accommodate the changes. Full details of the changes at a high level can be found below, along with a short guide on how to review this PR.

### BenchmarkVCFs

- Replaced fully with `SimpleBenchmark`.
- Updated to allow for multiple constant columns to be added to a run, e.g. further experiment parameters like coverage, etc. Used in the revised FE pipeline.

### FindSamplesAndBenchmark

- Fully replaced by `FindSamplesAndSimpleBenchmark`, which was a streamlined version of the old wrapper around `BenchmarkVCFs` to perform fingerprinting to match samples and then benchmark. 
- Removed old inputs JSON since no longer applicable.

### CompareBenchmarks

- For now this was just patched to handle renaming `Interval` column to `Stratifier` for backwards compatibility.
- In general, this could probably be replaced with a preprocessing script in the Google sheets export notebook with a few lines of pandas. 

### FunctionalEquivalence

- Revamp the top level structure of the WDL using some structs to help group certain inputs together and avoid some index arithmetic.
- Simplify some of the scatter blocks to require less nesting.
- Reimplement the `FEEvaluation` script overall using some simplified pandas code.
- Patch the `F1Evaluation` and `PlotROCs` tasks to handle the new benchmark formats.
- Allow `F1Evaluation` to take `plot_qual_limit` input from user. This fixes issue #80 
- Remove `RenameSummary` task and replace information in file names with actual data in tables. This fixes issue #49 
- Removed `scripts` directory since it contained Python/R versions of scripts that were outdated compared to those in the WDL directly.
- Refactor `F1Evaluation` using pandas to replace PR #110 

### Misc

- Removed `BenchmarkVCFs/CreateIGVSession.wdl` since this has been superseded by the utility WDL equivalent.
- `BenchmarkVCFs/shinyApp` probably no longer useful... should we delete in favor of BenchmarkBoard?
- Patch `MatchFingerprints` utility to localize VCF files to avoid Picard bug (see comments for details) for use with our RPays truth data in benchmarking
- Updated a bunch of documentation to reflect the above changes.

## How to Review

Although the number of lines of code changed is daunting, the actual content of the PR is much smaller because:
- `SimpleBenchmark` has already been reviewed, and is just being copy/pasted over to `BenchmarkVCFs`, so this file doesn't need to be reviewed again (unless you want to; there are some very small tweaks mentioned above).
- `FindSamplesAndSimpleBenchmark` has also been reviewed previously and is just replacing `FindSamplesAndBenchmark`.

So the main checklist of things to review is:
- [ ] `FunctionalEquivalence` updates: mainly the `FEEvaluation` and `F1Evaluation` tasks have been reimplemented to fit the new shape of the data, and make some small improvements to the readability (I hope). The rest of the tool is more or less the same beyond small refactors.
- [ ] `CompareBenchmarks` updates: these are pretty superficial and just to fit the new WDL format.
- [ ] `MatchFingerprints` updates: had a few changes to fix some bugs in testing that slipped by the initial FSASB review.
- [ ] Docs updates: all the docs have been updated!
- [ ] A number of files were deleted since they seem to be no longer relevant or are redundant with functionality elsewhere. A thumbs up that these are no longer needed would be appreciated (reasons given in the changelog above).

## What's left to do before merging?

Before actually merging, we still need to:
- [ ] Make announcements that the main branch will break dockstore imports for these tools
- [ ] Update the BenchmarkVCFs/FE pipeline to handle the vcfeval changes once [this PR](https://github.com/RealTimeGenomics/rtg-tools/pull/13) gets released to complete the `F1Evaluation` overhaul
- [ ] Double check any GATK tests using FE won't break with the merge
- [ ] Make sure that the FE featured workspace gets updated relatively soon after this merge happens


kachulis:
Thanks @rickymagner!  Can you rebase onto main so we can get the tests to run?

Could you also reach out to David to find out if any gatk tests are using FE from our main branch, as opposed to a tag?  If tests pass and there's no direct effect on the gatk tests from these changes, then I think this is good to go.

rickymagner:
Thanks, I'll work on the rebasing once I finish testing the last feature I just tried to add to make FE updated. And I'll message David now to see about GATK tests

rickymagner:
I think this is ready for final review @kachulis @michaelgatzen or whoever is able to review. I rebased on the latest main and fixed the last bit of making the ROC plots work with the new vcfeval PR. I tested it on the FE featured workspace workflows, with some data shown below for before/after. Also, the tests are currently failing since they're hooked up for the old pipeline, so either I could: delete them for now, and make adding tests a to do; or try to copy the data into the current pipelines and regenerate the test outputs to replace those. Let me know what you'd prefer.

## Before/After Comparison
In the FE workspace, there's a GATK 3 vs 4 and DRAGEN vs DRAGEN-GATK FE example pipeline. I ran those with the old main vs this current branch to compare the outputs. Remember that because some of the minor benchmarking conventions have changed, the numbers won't align perfectly for the summary stats. The ROC stats however should be essentially identical (there might be some weird boundary noise if variants are close to the edge of the special bed files). I extended the curves to the lowest amounts on the x-axis: not sure if the old behavior was intentional or if this is a fixed bug. Also the scaling for the F1 plots in the first comparison looks better in the new version (to me) and I have no idea why it's different now. Maybe the matplotlib versions or something? Also some labels for FE violations in this comparison changed (for the F1 plots) I think for the "better": SNP LCR and INDEL HCR. I think there was a bug in the previous computation (I don't remember the details anymore), but looking at e.g. the INDEL HCR lines, it definitely looks like it shouldn't be functionally un-equivalent there.

I can share the workspace I tested this in if you want or otherwise help you do your own testing.

### GATK3 vs GATK4

#### FE Plots

Before:
image

After:
image

Animated to see changes quickly:

https://github.com/user-attachments/assets/d6731199-8179-424d-90e2-d056903659c1

#### F1 Plots

Before:
image

After:
image

Animated:

https://github.com/user-attachments/assets/f4926829-1cdb-4926-85e7-fe5fa4f8e971


### DRAGEN vs DRAGEN-GATK

#### FE Plots

Before:
image

After:
image

Animated to see the changes quickly:

https://github.com/user-attachments/assets/8dbf862d-6a35-414e-944c-0485340b0b04

#### F1 Plots

Before:
image

After:
image

Animated:

https://github.com/user-attachments/assets/ae2c66c1-9bd8-4c4b-badf-df3b036a5aa6

BenchmarkVCFs/CompareBenchmarks.wdl
```diff
@@ -388,7 +394,7 @@ def main(sample_ids, configurations, summaries, stratifiers, order_of_samples, o
         sample_data['configuration'] = configurations[i]
         samples_data.append(sample_data)
     data = pd.concat(samples_data)
-    data = data.fillna({'Stratifier': 'all'})
+    data = data.fillna({'Stratifier': 'WholeGenome'})
```
michaelgatzen, line 397: This is getting very technical but I've run this pipeline on WES data in the past and then this label wouldn't be very accurate. Maybe AllRegions or something like that?

BenchmarkVCFs/CompareBenchmarks.wdl
```diff
@@ -403,7 +409,7 @@ def main(sample_ids, configurations, summaries, stratifiers, order_of_samples, o
     if stratifiers is None:
         stratifiers = data['Stratifier'].unique()
     else:
-        stratifiers = ['all'] + stratifiers
+        stratifiers = ['WholeGenome'] + stratifiers
```
michaelgatzen, line 412: If you end up changing the label above, also change it here. There are a few more in FEEvaluation and F1Evaluation

FunctionalEquivalence/FunctionalEquivalence.wdl
```diff
@@ -19,301 +36,257 @@ workflow FunctionalEquivalence {
...
+        Boolean passing_only = true
+        Boolean require_matching_genotypes = true
+        Boolean enable_ref_overlap = false
+
         Boolean signed_difference = false
 
-        String gatkTag="4.0.11.0"
-        Boolean passingOnly=true
-        Boolean requireMatchingGenotypes=true
-        String vcfScoreField = "QUAL"
```
michaelgatzen, line 41: Could we also port the score field into the new arguments?

Utilities/Dockers/VCFEval/Dockerfile_tmp
michaelgatzen, line 1: Just making sure that you need this file, the `_tmp` might suggest that you didn't mean to commit this

michaelgatzen (commented):
I just looked through the Functional Equivalence code. Amazing work, I really like the new structure with the data tables, that's so much nicer than whatever the hell I fabricated 4 years ago.

Did you ever test FE with an input of multiple datasets? It looks like it's doing the right thing but it might be worth checking that that's the case.

Thanks a lot for working through my code, this is great!

FunctionalEquivalence/FunctionalEquivalence.wdl
rickymagner: Good idea, just put the `score_field` as an input.

Utilities/Dockers/VCFEval/Dockerfile_tmp
rickymagner: Yeah, just to have this written somewhere in case people forget: the `_tmp` is intentional since it's a docker image built using just a public commit from the `rtg-tools` repo, rather than a formal release (which is supposedly coming sometime!). As some comments there suggest, that image should be replaced by one built from a formal tag with the new changes once it actually exists.

rickymagner:
Just finished the next/possibly last round of updates for this PR required to get it finished. In particular, the main highlights are:

- Updated the BenchmarkAndCompareVCFs workflow a bit. After chatting with Michael, we decided to remove the compute GC plots task since it was a bit niche, and I updated the tests accordingly. Please take a look and see if they're reasonable. Here's a screenshot of what the main outputs look like in Google sheets for the test data:

image

I also removed the old CompareBenchmark tests to be subsumed by these, so it's easier to recreate new tests in the future.

- I verified/fixed the multiple dataset mode for Functional Equivalence. Here's an example of how it looks using two copies of the same dataset.

https://github.com/user-attachments/assets/ac690810-1a01-4971-b224-fb01495d6475

Assuming the tests pass now, I think it's ready to merge unless there's more feedback.

kachulis:
this is awesome!  A couple of questions:

1.  it looks like the error bars in the FE plots have changed slightly (they look smaller in the "new" version).  Is this intended/expected?
2. I get that the gc plot implementation is a bit blah, but being able to create those performance vs gc plots is actually super useful, imo.  I'd almost rather leave the option in there, and then think about a better way to implement it in the future instead of removing it entirely.
3.  Didn't the CompareBenchmarks output used to have a bunch of pretty conditional formatting?  or is that something that was added separately (might be a question to @michaelgatzen)

rickymagner:
1. I think some amount of wiggle is expected for the FE plots since that's affected by the changes in how we count e.g. SNPs/INDELs using bcftools now. So the few edge cases that are handled differently should account for that noise, I think.

2. I talked with Michael about this yesterday and he said it's unlikely anyone is currently using it, and I think it's a bit brittle to infer behavior based on label prefixes. I have a (not yet pushed) utility for adding some GC-related annotations to VCFs that could be a partial replacement for this functionality, as you could take the vcfeval output with labels (just pushed a commit now to expose this in the final outputs) and then annotate with GC metadata to make a refined stratification by precise GC content, etc. I think in the longterm that's a better strategy that could be wrapped in another workflow to achieve similar/better functionality.

3. I think the pretty output formatting to have a table with spaces between some columns has remained untouched here, right? Or do you mean something else?

michaelgatzen:
1. Yeah I don't have big concerns about the error bars changing
2. We did talk about it and it's not anything that can't be achieved with a small Terra notebook. It is a [documented feature](https://github.com/broadinstitute/palantir-workflows/blob/main/BenchmarkVCFs/README_CompareBenchmarks.md#generating-plots-for-different-gc-content) that arguably is kind of brittle as Ricky said, I'm fine either way. I don't know if a lot of people use the plots that this workflow generates, I mostly added it as a convenience for me doing the sequencer evaluation 2 years ago, when I had to make these plots all the time
3. Correct, the formatting is done in a Google Colab notebook for permissions reasons, as documented [here](https://github.com/broadinstitute/palantir-workflows/blob/main/BenchmarkVCFs/README_CompareBenchmarks.md#notebook-for-exporting-results-to-google-sheets). If the output format of the `comparison_csv` doesn't change then that notebook will continue to work

kachulis (approved):
sounds good, thanks for the responses to my questions!  

I'm happy with merging this now, thanks for the awesome work @rickymagner!

rickymagner:
Thanks everyone for the great feedback! Let me make a Slack announcement we're going to merge this soon (probably tomorrow) so anyone e.g. in TAG using it has a last chance to pin their versions if they're in the middle of a project and imported off of the main branch.

rickymagner:
Just made a change to tell `watt` to ignore the `vcfeval` output VCF, since it caused the tests to fail due to some stochastic data (like date, cromwell filesystem dirs, etc) being stored in the header, so it's hopeless to be able to get an exact match. Longer term we could consider some updates to `watt` itself to better handle doing a more intelligent VCF comparison, but for now I think it's easiest to ignore the output. Hopefully the tests will now pass!

PR #168, opened 2024-04-11
Remove cram localization in GLIMPSE phase

kachulis:
https://github.com/odelaneau/GLIMPSE/pull/208 allows cram index to be specified when running GLIMPSE, so that ref caching and partial cram localization are no longer necessary.


kachulis:
replaced by #170

PR #169, opened 2024-04-12
AXIN hotfix

kachulis:


michaelgatzen (approved):
Approving so you can merge it anytime, but it would be great to keep the Glimpse2ImputationInBatches.wdl in sync by making `cram_indices` optional in that WDL, everything else should be able to stay the same

kachulis:
@michaelgatzen agreed, will update Glimpse2ImputationInBatches.wdl as part of #168

PR #173, opened 2024-05-14
Glimpse/htslib retries & bug fix, and index specification

kachulis:
A few changes here.  

- implements use of `##idx##` to specify cram index
- updates glimpse docker for retries and to fix bug which can lead to partial reading of crams from cloud storage (see odelaneau/GLIMPSE#212)
- adds coverage metrics as additional output to try to catch any potential issues like this in future
- check for "EOF marker missing" message from htslib which occurs with one remaining edge case read failure which cannot otherwise be caught
- some updates to docker building to be able to build on m1 mac (for linux/amd64)

might worth adding a check on the coverage metrics to Glimpse2CheckQC.  I can do that in a separate pr.


GlimpseImputationPipeline/Glimpse2Imputation.wdl
```diff
@@ -165,28 +171,18 @@ task GlimpsePhase {
         cram_index_paths=( ~{sep=" " cram_indices} )
         sample_ids=( ~{sep=" " sample_ids} )
 
-        if ~{if defined(cram_indices) then "true" else "false"}; then
-            seq_cache_populate.pl -root ./ref/cache ~{fasta}
-            export REF_PATH=:
-            export REF_CACHE=./ref/cache/%2s/%2s/%s
-        
-            chunk_region=$(echo "~{reference_chunk}"|sed 's/^.*chr/chr/'|sed 's/\.bin//'|sed 's/_/:/1'|sed 's/_/-/1')
+        duplicate_cram_filenames=$(printf "%s\n" "${cram_paths[@]}" | xargs -I {} basename {} | sort | uniq -d)
```
michaelgatzen, line 174: Given that we now explicitly define the indices, are duplicate basenames still an issue? Are the index file still localized?

(It's possible that you're going to answer this question in your deep dive in 30 minutes, haha)

GlimpseImputationPipeline/Glimpse2Imputation.wdl
```diff
@@ -209,7 +205,11 @@ task GlimpsePhase {
             cmd="$cmd --checkpoint-file-in checkpoint.bin" 
         fi
 
-        eval $cmd
+
+
+        #check for read error which corresponds exactly to end of cram/bam block.  
+        #This currently triggers a warning message from htslib, but doesn't return any error
+        eval $cmd |& tee >(if grep -q "EOF marker is absent"; then echo "An input file appears to be corrupted.  Please fix before retrying."; exit 1; fi;)
```
michaelgatzen, line 212: Is it obvious how the input file would be corrupted? Like would it fail `samtools quickcheck` for example? I'm just trying to imagine coming across this error message, hoping to know what to do in that case

GlimpseImputationPipeline/Glimpse2Imputation.wdl
```diff
@@ -429,3 +430,46 @@ task SelectResourceParameters {
...
+        paste chunk_col.txt cov_file.txt > coverage_metrics.txt
+
+    >>>
+
+    runtime {
+        docker: "ubuntu:24.04"
+    }
+
+    output {
+        File coverage_metrics="coverage_metrics.txt"
```
michaelgatzen, line 473: I tried to name all output files with the sample name (or `basename`) in the filename, in order to make metrics storage and debugging slightly easier. If you want you can implement that here by passing `basename` to this task and changing the output file name to `~{basename}.coverage_metrics.txt` (also in line 464 of course)

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env zsh
```
michaelgatzen, line 1: Is this required? I don't know if zsh is available on all systems

GlimpseImputationPipeline/Glimpse2Imputation.wdl
kachulis: the indices are still optional inputs, so I think we still want to check for duplicates.

michaelgatzen: At this point, do we just want to require the indices? I don't know what happens when no index is provided (is one created? Or is just the file parsed until it encounters the region?), but either way it feels inefficient to do that 800 times in parallel

GlimpseImputationPipeline/Glimpse2Imputation.wdl
kachulis: http server returning an error that occurs at exactly the wrong time in the streaming of the cram.  I've never seen it, and I think it'd be super rare.  But I can't convince myself it's not possible.

I can adjust the message to be clearer and mention this possible cause as well.

GlimpseImputationPipeline/Glimpse2Imputation.wdl
kachulis: good call, will add basename

GlimpseImputationPipeline/glimpse_docker/build_base_and_extension_docker.sh
kachulis: yeah, I'm not sure about this.  when the environment was bash the build was failing in weird ways I couldn't explain and making it zsh fixed the problem.  Probably more an issue with my system than anything else, so I'm fine with switching it back to bash and trying to deal with it on my side.

GlimpseImputationPipeline/Glimpse2Imputation.wdl
michaelgatzen: Yeah I see 3 options here, all of them are fine
1) Leave it as is
1) Add a check in bash that, in case of CRAM input, verifies if indices are defined
1) Remove VCF input option and make CRAM and CRAI required

I'm going to approve this PR, feel free to add any of these changes or not.

michaelgatzen (approved):
I added one more comment in the cram index discussion thread

PR #176, opened 2024-05-29
Use watt for tests

kachulis:
requires rickymagner/watt#2


.circleci/config.yml
```diff
@@ -9,59 +9,45 @@ executors:
...
 
-  install-make:
-    description: "Install make"
+  install-watt:
+    description: "Install watt"
     steps:
       - run: |
-            apt-get update
-            apt-get install -y make
+            git clone -b ck_ignore_output https://github.com/kachulis/watt.git ~/watt
```
kachulis, line 23: change this once rickymagner/watt#2 is merged

rickymagner: If you're all set making tweaks, I'd also rather make a tagged release for Watt to use here

.circleci/config.yml
```diff
@@ -9,59 +9,45 @@ executors:
...
-      CROMWELL_JAR=/home/circleci/project/cromwell-84.jar
     steps:
       - checkout
       - get-cromwell-jar
-      - run: make test
-      - store_artifacts:
-          path: /home/circleci/project/logs
+      - install-watt
+      - run: |
+            python ~/watt/watt.py -e cromwell-84.jar -c test/watt_config.yml -p 2
```
rickymagner, line 40: Are you sure you only want to use `-p 2` here? In practice it seemed to be useful going up to `$(nproc)` or so.

.circleci/config.yml
kachulis: sounds good to me

kachulis: a couple other potential changes on my mind, but I think this is a good place for a first tagged release

test/watt_config.yml
```diff
@@ -0,0 +1,49 @@
...
+  workflow_name: "AggregatePRSResults"
+  test_name: "Base"
+  path: "/ImputationPipeline/AggregatePRSResults.wdl"
+  test_inputs: "/test/AggregatePRSResults/test_inputs.json"
+  expected_outputs: "/test/AggregatePRSResults/test_outputs.json"
+
+benchmark_vcfs_ignore_filters_test:
+  workflow_name: "BenchmarkVCFs"
+  test_name: "IgnoreFilters"
+  path: "/BenchmarkVCFs/BenchmarkVCFs.wdl"
```
rickymagner, line 11: Have you tested running these locally with this directory setup? I think Watt is configured so that the location of the `yml` is the "root" when resolving these sorts of paths, so this file should live one directory up I think (unless I changed it to look up until it finds the root of the repo, which I was thinking about but think I skipped).

test/watt_config.yml
```diff
@@ -0,0 +1,49 @@
...
+
+compare_benchmarks_test:
+  workflow_name: "CompareBenchmarks"
+  test_name: "Base"
+  path: "/BenchmarkVCFs/CompareBenchmarks.wdl"
+  test_inputs: "/test/CompareBenchmarks/test_base_inputs.json"
+  expected_outputs: "/test/CompareBenchmarks/test_base_outputs.json"
+
+
+## proposed new format
```
rickymagner, line 37: This is not a bad idea. I'd be up for changing Watt to do something like this. I'm not a fan of having the empty `-` lines, so maybe something like:

```
workflow_name:
  path: "/my/path"
  tests:
    test_name:
      test_inputs: "/here/they/are"
      expected_outputs: "expected/out"
```
I think this is all legal yaml, no?

.circleci/config.yml
```diff
@@ -9,59 +9,45 @@ executors:
...
       - store_artifacts:
-          path: /home/circleci/project/workflow_logs
+          path: watt_logs
       - store_artifacts:
-          path: /home/circleci/project/call_logs
-
+          path: cromwell-executions
```
rickymagner, line 44: What effect does changing this have on finding/reading logs when tests fail in CI?

rickymagner (commented):
Made a few comments. Thanks for doing this!

test/watt_config.yml
kachulis: It actually looks like you did implement a [search upwards for a root of a github repo](https://github.com/rickymagner/watt/blob/c01b17e6e1f2163802caf511a3c73d5afa365f1d/watt.py#L37) :)

however, the tests do still need to be run from the palantir-workflows root, because the paths in the input and expected json files are still interpreted (both by watt an cromwell) as relative to the current directory.  Would be pretty simple to connect that up to the `resolve_relative_path` functionality that is already there, but I don't think worth holding up.

opened [an issue](https://github.com/rickymagner/watt/issues/3) so we can maybe remember to do this is anyone has time

test/watt_config.yml
kachulis: oh, yeah, I thought this might just be a little easier to parse and keep organized this way.

yeah, keeping everything as key-value mappings instead of using block sequences is cleaner, I agree.

Do you have a preference on reworking the format before/after the initial release?

.circleci/config.yml
kachulis: should be somewhat easier, in that you have the entire cromwell execution directory now.  

that said, there isn't a ton of guidance when a test fails as to where to look for the failures (particularly if there is more than on test of the same workflow, figuring out which one to look in for the failure seems to be completely guess and check, atm).  I think that's something we could improve in watt.

test/watt_config.yml
rickymagner: Thanks for opening the issue. It sounds like the current limitation is just that the JSONs themselves have to use paths relative to the runtime directory, but the config file uses the repo root as `/`. I think I agree then that it would make sense to parse the JSON paths as also being relative to the repo root.

test/watt_config.yml
rickymagner: I would say it's unlikely I'll be able to get around to doing this refactoring before later next week, so if you prefer to get the tests updated here ASAP I'd be OK with merging this PR using the old format (with a v0.1 tag on Watt), and then update Watt to this convention for a v0.2 sometime in the near future.

.circleci/config.yml
rickymagner: I was more confused about the change from a file path to a "local" path in the new commit. But yeah, for Watt it will tell you which tests failed and for what reason/outputs. Generally you shouldn't need the cromwell logs unless you have a failed test because the workflow failed to finish unexpectedly. So the Watt logs would generally be the place to look to see the test results summary, and then you can debug from there.

rickymagner:
The config refactor looks good to me, but it seems that `BenchmarkVCFs/base` failed because Cromwell didn't finish for some reason. Do you know if it's possible to look at the VM and see the cromwell logs? It might get messy to try to print that to stdout in the logs (because of the concurrency), but that's an option. Of course running locally is probably the easiest quick fix and then check those logs.

kachulis:
cromwell logs are saved as artifacts, so I will take a look

rickymagner (approved):
LGTM. Thanks for setting this up!

PR #179, opened 2024-06-13
Testing overhaul

kachulis:
two major changes

1. moves CI testing from circleci to github actions
2. changes wdl tests from running with local backend to running with GCP Batch backend

requires rickymagner/watt#7


.github/actions/install-watt/action.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: 'installs WATT'
+inputs:
+  WATT_VERSION:  # id of input
+    description: 'WATT tag to use'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Clone WATT
+      run: git clone -b ${{ inputs.WATT_VERSION }} --depth 1 https://github.com/kachulis/watt.git ~/watt
```
kachulis, line 11: change once watt pr is in

.github/workflows/run_tests.yaml
```diff
@@ -0,0 +1,43 @@
...
+        uses: 'google-github-actions/auth@v2'
+        with:
+          workload_identity_provider: 'projects/1038484894585/locations/global/workloadIdentityPools/github-wi-pool/providers/github-wi-provider'
+          service_account: palantir-workflows-service-acc@broad-dsde-methods.iam.gserviceaccount.com
+      - uses: ./.github/actions/install-cromwell
+        with:
+          CROMWELL_VERSION: 87
+      - uses: ./.github/actions/install-watt
+        with:
+          WATT_VERSION: ck_gcp
```
kachulis, line 36: change once watt pr is in

README.md
```diff
@@ -1,6 +1,6 @@
 # PALANTIR-WORKFLOWS
 
-[![CircleCI](https://circleci.com/gh/broadinstitute/palantir-workflows.svg?style=svg)](https://circleci.com/gh/broadinstitute/palantir-workflows)
+[![Test Status](https://github.com/broadinstitute/palantir-workflows/actions/workflows/run_tests.yaml/badge.svg?branch=ck_github_actions_google_cloud)](https://github.com/broadinstitute/palantir-workflows/actions/workflows/run_tests.yaml/badge.svg?branch=ck_github_actions_google_cloud)
```
kachulis, line 3: branch --> main

kachulis:
CircleCI tests fail because CircleCI config yaml is removed, but CircleCI is still active for this project.  Once this pr is merged we can deactivate circlci for this repo, so that failure can be safely ignored, I think.

test/simple_cromwell.conf
```diff
@@ -0,0 +1,71 @@
...
+
+          parallel-composite-upload-threshold="150M"
+        }
+
+        filesystems {
+          gcs {
+            auth = "application-default"
+            project = "broad-dsde-methods"
+
+            caching {
```
rickymagner, line 43: For caching, this will look at the `test-execution` bucket for previous runs of the test, correct? Is there a way to make sure that bucket stays "clean" with just one copy to call-cache from and not build up a bunch of clutter over time from each PR? Does it have some TTL policy to help with this?

rickymagner (changes requested):
Looks good overall, thanks. It seems like the biggest changes are moving a bunch of files to the cloud that watt can pull from to run the Cromwell tests in GHA. I'll keep a "Request Changes" on this to make sure we update the correct branches you left comments for after the watt PR is merged

test/simple_cromwell.conf
kachulis: I think there are a few options here, though I don't think any of them are perfect.  We could set `duplication-strategy` to `reference`, which would reduce call-caching related clutter, however would not play well with TTL, since call-caching would break every few months when the bucket was cleaned up.  At the moment, I'm leaving this as `copy`, but with TTL of 90 days, so things will get cluttered, but there will be a consistent cleanup as well.  I think once we see how much data actually sits in this bucket at steady state we can figure out the right settings.

The other thing to note is that call-caching is "on" in name only currently, since I haven't actually set up an external database to cache from.  I think that could be a nice addition in the future, but again I think it's worth getting an idea what the compute costs associated with this testing actually are, since a cloud sql instance on its own looks like it would cost $10-20 per day.  Would be nice for the speedup, but it's possible we should think of it more as a time saver than a money saver.

rickymagner: $10-$20 per day still seems like a lot for a repo that doesn't have that many PRs per day typically. Does the `copy` setting just look inside the bucket for a previous run and then cache based on that? I think doing a pilot with 90 days TTL sounds reasonable to get started.

kachulis: The cache lookup is based on a database, which is either in memory, or an external database.  When the database in in memory (as in this default case), it essentially means that it can't access info about any workflows from previous runs of cromwell.  

Once it finds info about a previous run to cache from, the `copy` vs `reference` option tells cromwell whether to copy the previous output files into a new location in the new execution directory, or just use the previous run outputs without creating a new copy of them.  At least that's my understanding, I've never actually used `reference`, so don't know if they're are any quirks in its behavior.

https://cromwell.readthedocs.io/en/stable/cromwell_features/CallCaching/

kachulis:
@rickymagner thanks for the review.  I have updated the watt version and branch reference in the readme, so I think this will be ready to merge once approved, (assuming tests pass).

rickymagner:
Looks good assuming tests pass, thanks!

rickymagner (approved):

PR #182, opened 2024-08-02
Update vcfeval docker image

rickymagner:
Just adding some extra documentation for the `vcfeval_docker:v1.1` used for an intern project this summer.

### DIFF
--- a/Utilities/Dockers/README.md
+++ b/Utilities/Dockers/README.md
@@ -134,7 +134,7 @@ includes some minimal Python tools (pandas) for data processing.
 * Usage: `rtg [COMMAND]`, etc.
 * Version Notes:
   * 1.0: Versions are `rtg` 3.12.1, `bedtools` 2.31.0, `bcftools` 1.16
-  * 1.1: Added htslib 1.20 (includes tabix and bgzip); bumped bcftools to 1.20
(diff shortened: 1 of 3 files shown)


kachulis (approved):

rickymagner:
Just rebased on main and made sure to rebuild + push the dockers to the cloud, since it seems they became out of sync with previous updates. I also took the moment to update the ubuntu base and bump the hts versions. If you have a second to double check the PR again, I'd appreciate it.

P.S. Technically when you updated some of the `Dockerfile` base ubuntu versions, they might've fallen out of sync with what's in the cloud. Hopefully this doesn't cause any issues, but good to remember.

kachulis (approved):
> P.S. Technically when you updated some of the Dockerfile base ubuntu versions, they might've fallen out of sync with what's in the cloud. Hopefully this doesn't cause any issues, but good to remember.

That is definitely the case.  I wonder if we might want to add automated building and pushing of any dockerfile that changes on a push to main.  That shouldn't be too difficult, though we'd need to figure out how to automatically bump tags and such things.

more complicated would also be automatically updating references to dockers in wdls when dockerfiles are updated/built/pushed.  I think it's doable, but would require a bit of thought on the best way to design (if it's even worth doing)

PR #183, opened 2024-08-06
Glimpse tests

kachulis:


GlimpseImputationPipeline/Glimpse2Imputation.wdl
```diff
@@ -80,8 +85,8 @@ workflow Glimpse2Imputation {
                 fasta_index = fasta_index,
                 preemptible = preemptible,
                 docker = docker,
-                cpu = select_first([safety_check_n_cpu, SelectResourceParameters.request_n_cpus]),
-                mem_gb = select_first([safety_check_memory_gb, SelectResourceParameters.memory_gb]),
+                cpu = select_first([cpu_phase, safety_check_n_cpu, SelectResourceParameters.request_n_cpus]),
```
rickymagner, line 88: Shouldn't the safety checks go first? Maybe I misunderstand what the purpose of the phase variables are.

test/simple_cromwell.conf
```diff
@@ -67,5 +67,79 @@ backend {
...
+        root = "gs://palantir-workflows-test-execution"
+
+        # Make the name of the backend used for call caching purposes insensitive to the PAPI version.
+        name-for-call-caching-purposes: PAPI
+
+        # Emit a warning if jobs last longer than this amount of time. This might indicate that something got stuck in PAPI.
+        slow-job-warning-time: 24 hours
+
+        genomics {
+          auth = "application-default"
```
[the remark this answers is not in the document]
kachulis: It uses [workload identity federation](https://cloud.google.com/iam/docs/workload-identity-federation) to authenticate [here](https://github.com/broadinstitute/palantir-workflows/blob/10fb7dc60fed1ed037fb5942261e8d21dd87e2e0/.github/workflows/run_tests.yaml#L31C7-L36C101), so we don't need to store any secrets in the repo.  The workload identity provider pool is managed by DevOps, I'll put some documentation on the wiki of how this is all set up.

GlimpseImputationPipeline/Glimpse2Imputation.wdl
kachulis: `safety_check_n_cpu` was intended since the parameters were being set algorithmically, so we don't want the user to be surprised by a request for a ton of resources.  My feeling with an input variable though is that the user shouldn't be surprised, since they have set the value in the first place.  So there's no need for the safety check in that scenario.  I could also see a scenario where for some reason we might want to push the resources higher than the safety check manually, and the wdl should allow for that.

Also, even without the safety check here, there is always the safety checks of quota that won't allow someone to ask for like a petabyte of ram.

rickymagner: I see, that make sense, thanks

rickymagner (approved):

PR #188, opened 2024-08-23
Scatter Glimpse Merge

kachulis:


GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -61,17 +119,169 @@ workflow Glimpse2MergeBatches {
...
+
+    python3 <<CODE
+    import glob, os
+    # Works around a JES limitation where multiples files with the same name overwrite each other when globbed
+    intervals = sorted(glob.glob("out/*/*.interval_list"))
+    for i, interval in enumerate(intervals):
+      (directory, filename) = os.path.split(interval)
+      newName = os.path.join(directory, str(i + 1).zfill(4) + filename)
+      os.rename(interval, newName)
+    print(len(intervals))
```
michaelgatzen, line 271: Works perfectly fine, it just feels a little nicer if it would be written to an actual file, don't you think?
```suggestion
    with open('num_intervals.txt', 'w') as num_intervals:
      num_intervals.write(f'{len(intervals)}\n')
```

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -146,6 +356,138 @@ task CountSamples { # really?
...
+        docker: docker_merge
+        disks: "local-disk " + disk_size_gb + " HDD"
+        memory: mem_gb + " GiB"
+        cpu: cpu
+        preemptible: preemptible
+    }
+}
+
+
 task MergeAndRecomputeAndAnnotate {
```
michaelgatzen, line 491: Can get rid of this entire task. The only other place this is referenced is in a comment which you could update as well then, depending on which option you choose about QC metrics being required or optional

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -23,33 +26,88 @@ workflow Glimpse2MergeBatches {
...
-                docker_merge = docker_merge
+                imputed_vcf = GatherVcfs.output_vcf,
+                imputed_vcf_index = GatherVcfs.output_vcf_index,
+                docker_count_samples = docker_count_samples
+        }
+
+        if (defined(qc_metrics)) {
+            call MergeQCMetrics {
+                input:
+                    qc_metrics = select_first([qc_metrics]),
```
michaelgatzen, line 107: This `select_first` will fail if qc_metrics is null. I see two options:
1. You can make QC metrics a required input to this WDL. It initially wasn't because we said that QC collection for imputation should be optional, but at this point we might as well always turn it on.
2. Make qc_metrics an optional input to the `MergeQCMetrics` task, and then handle the null case there differently (see comment for line 472)

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
```diff
@@ -146,6 +356,138 @@ task CountSamples { # really?
...
+    }
+
+    command <<<
+        set -xeuo pipefail
+        
+
+        cat <<EOF > script.py
+    import pandas as pd
+    qc_metrics = ['~{sep="', '" select_first([qc_metrics, []])}']
+    merged_qc_metrics = pd.concat([pd.read_csv(qc_metric, sep='\t') for qc_metric in qc_metrics])
```
michaelgatzen, line 472: You're seemingly handling the null case in the line above, but `pd.concat([])` will raise a ValueError. If we want to handle the null case then the output should be optional and line 472 and 473 should only be executed if qc_metrics is not null.

michaelgatzen (changes requested):
Looks great overall, I just have a few comments

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
kachulis: yup, agreed

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
kachulis: isn't this fine, because we'll only get into this block if qc_metrics is not null?

that said, I think your point about just making qc collection required makes sense, so going to take that route.

GlimpseImputationPipeline/Glimpse2MergeBatches.wdl
kachulis: just making qc_metrics required to avoid all this

kachulis:
thanks for the review @michaelgatzen, back to you

michaelgatzen (approved):

michaelgatzen (changes requested):
Looks like the test is failing

kachulis:
@michaelgatzen tests are now passing :)

michaelgatzen (approved):

PR #191, opened 2024-09-06
More testing and linting

kachulis:


.github/workflows/run_tests.yaml
```diff
@@ -77,4 +77,26 @@ jobs:
...
+          python-version: '3.9'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+      - name: Lint with Ruff
+        run: |
+          pip install ruff
+          ruff check --output-format=github .
+      - name: ScoreBGE Unit tests
+        run: |
```
rickymagner, line 98: Does this mean that linting + testing this particular test are wrapped up in the same GH action? It might be worth splitting them just for conceptual clarity.

rickymagner (approved):
I think this looks good overall. Just left one comment that I leave up to you to decide on

.github/workflows/run_tests.yaml
kachulis: good call, I will split them up

PR #199, opened 2024-11-07
Update BenchmarkSVs workflow(s) with some new features

rickymagner:
This PR includes a bunch of refactoring, improvements, and new features for the process of benchmarking SV VCFs. At a high level, this includes:

- Allow for the option to use `truvari refine` and `truvari ga4gh` for collapsing (or "harmonizing") similar events in truth/query down to one event to try to improve benchmarking statistics where calls might get mismatched due to extreme fuzziness in the calling step. This includes an alignment step using `mafft` as outlined in the `truvari` documentation on the process. The docker image is updated to include newer versions of truvari as well as mafft. This option requires the input files to be phased.
- Splits multiallelic sites before running truvari since the tool expects this (previous this would be expected for the users to preprocess in this way, but this adds a quick convenience).
- The QC tasks were split out into a separate workflow to streamline the benchmarking vs qc/counting tasks. Some common tasks were moved to a third file and are imported.
- Dockstore yml has been updated to be able to import both workflows there.
- The benchmarking stats also include a breakdown by HET vs HOM sites (and both together).

### DIFF
--- a/.dockstore.yml
+++ b/.dockstore.yml
@@ -144,3 +144,9 @@ workflows:
   - name: ScoreBGE
     subclass: WDL
     primaryDescriptorPath: /ImputationPipeline/ScoreBGE/ScoreBGE.wdl
+  - name: BenchmarkSVs
+    subclass: WDL
+    primaryDescriptorPath: /BenchmarkVCFs/BenchmarkSVs.wdl
+  - name: QcSVs
+    subclass: WDL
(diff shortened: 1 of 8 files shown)


kachulis (approved):
looks good!  only question is whether you want to add any tests.

rickymagner:
I'll put tests on the infinite to-do list and get this merged for now. Thanks for checking!

PR #202, opened 2025-02-04
Dockerfile tests and cleanup imputation/PRS

kachulis:
This PR does a few things.  It looks like a lot of changed files/lines, but the majority of that is just removing the minimac imputation pipelines, and renaming the `ImputationPipeline` directory to `PRS`.

###  1) Adds tests to build all Dockerfiles in order to maintain working Dockerfiles 

- PRs : all changed Dockerfiles are built
- push to main : all Dockerfiles are built
- once per week : all Dockerfiles are built (on main)

### 3) Removes minimac imputation pipeline

- This pipeline was long ago migrated to [warp](https://github.com/broadinstitute/warp/blob/develop/pipelines/broad/arrays/imputation/Imputation.wdl)
- A number of associated dockers were failing to build.  Rather than fix them for a pipeline we no longer maintain in this location, I would prefer to delete this version and keep only the maintained version in warp.

### 4) Rename `ImputationPipeline` directory to `PRS`

- Given the removal of the minimac impuation pipeline, that directory is now entirely PRS pipelines and associated utilities.  Makes sense to change the name to finally reflect what is actually there.


LongReadRNABenchmark/cupcake_docker/Dockerfile
```diff
@@ -51,6 +51,8 @@ WORKDIR $SRC
 
 COPY remove_fastq_duplicates.py $SRC/remove_fastq_duplicates.py
 
+ENV SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True
```
rickymagner, line 54: This feels like the sklearn version should also be pinned somewhere in the build, right?

.github/workflows/test_dockerfiles.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - uses: actions/checkout@v4
+      -
+        name: Check if Dockerfile has changed
+        id: check-if-changed
+        if: ${{ github.ref != 'refs/heads/main' }}
+        uses: tj-actions/changed-files@v44
+        with:
+          files: '**/GlimpseImputationPipeline/glimpse_docker/Dockerfile'
+      -
+        name: Build main GLIMPSE Docker
```
rickymagner, line 84: What's the purpose of testing the build of the docker image? Presumably to test any changes made in the pipeline, one would have to build the docker, upload it, and run the WDL already, right?

PRS/plink2_docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# default ubuntu version: 20.04
+ARG UBUNTU_VERSION=20.04
```
rickymagner, line 2: Why use a much older image for a new docker image? I think 24.04 is the latest LTS image. Unless any of the packages break in newer versions?

Utilities/Dockers/Samtools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.10
+FROM ubuntu:22.04
```
rickymagner, line 1: Should this also bump to 24.04?

rickymagner (changes requested):
This looks good overall. Just made some comments that should be addressed particularly about some docker image versions.

LongReadRNABenchmark/cupcake_docker/Dockerfile
kachulis: agreed, but the issue is actually that sklearn really shouldn't be installed, the correct pypi package is scikit-learn.  sklearn is kept around [to prevent malicious actors from using it](https://pypi.org/project/sklearn/).  As far as I can tell from that link, there is no version of sklearn that can be installed without setting this env variable.  The "right" way to fix this is to insall scikit-learn instead, but since sklearn is set as a requirement in the cupcake setup.py, I think this would require making a change to cupcake itself, which doesn't seem worth it since the tool itself is not longer maintained.

.github/workflows/test_dockerfiles.yaml
kachulis: I'm not sure I follow?  All of these docker tests are just to make sure they build, not to test whether the build work correctly in the pipelines.  The one difference here is that we need to build the glimpse tool docker on it's own and then build the glimpse pipeline docker from that.  I believe the reason for doing that was basically to not have to duplicate the glimpse tool docker in our pipeline docker, but still be able to add things to it.

kachulis:
@rickymagner, updated the docker versions, back to you!

LongReadRNABenchmark/cupcake_docker/Dockerfile
rickymagner: I see, that makes sense. Maybe can you add a comment to the Dockerfile just quickly explaining this for future excavators? Or even just a link back to this comment would suffice

rickymagner (approved):
I had a few discussion points above, but I think the changes are good and ready to merge when you are.

.github/workflows/test_dockerfiles.yaml
kachulis: I agree testing that an updated docker actually does what you want would be even better, but I think that would require automatic build-push (easy) and associated wdl update (more complex but intriguing).  For the moment I'm just thinking of this as both a check to catch obvious mistakes, and a check against external changes breaking our Dockerfiles.

PR #203, opened 2025-02-11
GLIMPSE2 improvements

michaelgatzen:
- Added `mem_scaling_factor_phase` for adjusting memory requirements when running in single sample mode
- Updated GLIMPSE2 docker to most recent version from official repo
- Removed monitoring log output, since Terra now has this feature built in
- Added default value for `interval_list` in `Glimpse2MergeBatches`
- Updated documentation in README
- Added automatic tagging in for docker generation script
- Had to rename the `die` function in the shell script to `end_program` because Github Copilot would refuse to do anything in that script

### DIFF
--- a/GlimpseImputationPipeline/Glimpse2Imputation.wdl
+++ b/GlimpseImputationPipeline/Glimpse2Imputation.wdl
@@ -23,12 +23,13 @@ workflow Glimpse2Imputation {
         Int? effective_population_size
         
         Int preemptible = 9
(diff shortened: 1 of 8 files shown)


michaelgatzen:
@kachulis I fixed the error message (that didn't actually stop me from running the workflow in Terra). Also from the reorganizing of the files the test for `ValidateScoring` failed, I changed the path in that WDL.

kachulis (approved):
Looks good!  The fix to the validation wdl also just went in in #204, so if you'll probably need to rebase to resolve that conflict.

PR #207, opened 2025-03-03
Wire Glimpse phasing runtime params into outer workflow

geoffjentry:
I told @kachulis that we'd made a small update to the glimpse wdl to wire the phasing cpu/mem into the outer WDL. This PR recapitulates that update.


kachulis (approved):
Looks good thanks @geoffjentry

PR #209, opened 2025-03-19
Always generate merged vcf, regardless of cnv qc status

kachulis:


michaelgatzen (approved):
If the short_variant_vcf_index really isn't needed then sure

PR #212, opened 2025-03-28
Move tests to GCPBatch

kachulis:
PAPIv2 will be [unavailable after July 8, 2025](https://cloud.google.com/life-sciences/docs/how-tos/migration).  We will need to migrate our tests to GCPBatch.

Leaving PAPIv2 option in cromwell config, but moving tests to run on GCPBatch.  We can remove PAPIv2 completely from the config if this change appears to be working fine for a while.


rickymagner (approved):
LGTM assuming the new tests work.

PR #221, opened 2025-05-15
Fix Python-Data-Slim Dockerfile

kachulis:
[`package_index` was recently removed from setuptools](https://github.com/pypa/setuptools/pull/4963#issuecomment-2851940964), but is used in firecloud `setup.py`, so installing firecloud in the docker fails.  

fix is to install older version of setuptools, and then force pip to use the installed version of setup tools with the [`--no-build-isolation`](https://github.com/pypa/setuptools/discussions/4354) option.


kachulis:
also adds a badge to the readme for Dockerfiles status

rickymagner (approved):
Looks good, thanks for fixing this. Hopefully one day Terra will have a new API wrapper in Python we can convert to, but for now this is good.

PR #224, opened 2025-07-23
BGESexEstimation script

michaelgatzen:


BGESexEstimation/BGESexEstimation.py
```diff
@@ -0,0 +1,57 @@
...
+        reader = csv.DictReader(file, fieldnames=['contig', 'num_reads', 'mean_coverage'])
+        for row in reader:
+            if row['contig'] == 'chrX':
+                covX = float(row['mean_coverage'])
+            elif row['contig'] == 'chrY':
+                covY = float(row['mean_coverage'])
+    return covX, covY
+
+def estimate_sex_ploidy(covX, covY):
+    lower_bound_xx = 0
```
kachulis, line 18: probably worth making these input parameters so that if we ever want to change them we don't need to change the source code of the script

PR #229, opened 2025-10-31
Update dockers for BenchmarkVCFs

rickymagner:
Hi folks, I was just running BenchmarkVCFs myself and noticed the dockers were a bit outdated when I was setting it up. In particular, there's an old ugly `vcfeval:v1.1-tmp` docker that is no longer necessary due to RTG formally releasing a new version back in May. I decided to go ahead and bump a bunch of other versions associated with the pipeline since I was rebuilding dockers.

WARNING: This PR is not ready to merge until someone in hydro.gen does the following:
1. Check out this branch so you can see the updated Dockerfiles.
2. For each of the `BCFTools`, `Python-Data-Slim` and `VCFEval` Dockerfiles, do the following:
  - Build the docker.
  - Push to the corresponding location in the DSDE Methods storage (e.g. `us.gcr.io/broad-dsde-methods/bcftools:v1.4`, `us.gcr.io/broad-dsde-methods/python-data-slim:1.1`, and `us.gcr.io/broad-dsde-methods/vcfeval_docker:v1.2` respectively). I kept the mistaken convention on not having a "v" for versioning the python-data-slim docker that happened for some reason, but do whatever you like there. Also at some point the Dockerfile for that one and README became out of sync so this should fix that as well.

After that it should be good to go.


kachulis (approved):
This looks good, thanks @rickymagner!  And sorry about the long delay getting to this?

I'm going to try to figure out how to get tests to run on this, and then will merge.

edit: can't figure out why actions didn't run.  I'm just going to merge and deal with failures if they happen.

PR #230, opened 2025-11-03
HPVDeepSeek v0-prerelease

kockan:
There will certainly be changes/improvements to this draft as we get closer to the initial launch. This PR is intended to provide an initial round of review and enable discussion on any choices made during the implementation


HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1281 @@
...
+            output_basename = output_basename
+    }
+
+    call FastQC as PostTrimmedFastQC {
+        input:
+            r1_fastq = TrimAndFilter.fastq1_trimmed,
+            r2_fastq = TrimAndFilter.fastq2_trimmed
+    }
+
+    call BwaMem as AlignReads {
```
kachulis: is it really necessary to perform this pre-consensus alignment and collect all of these pre-consensus metrics?  or could you go straight to calling molecular consensus and skip all of this?

kockan:
@kachulis Ready for a round of review

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+        String bait_set_name
+        String read_group_id
+        String read_group_sample_name
+        String read_group_library_name = "LB_DEFAULT"
+        String read_group_platform = "ILLUMINA"
+        String read_group_platform_unit = "PU_DEFAULT"
+        String read_group_description = "DS_DEFAULT"
+        String read_structure = "3M2S+T"
+
+        # HPVDeepSeekSomaticVariantCalling inputs
```
kachulis, line 37: Should this be part of the initial release?

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+        File realignment_index_bundle
+        String mapping_filter_python_script = "/usr/filter_alt_ref_positions.py"
+        File blastdb_nhr
+        File blastdb_nin
+        File blastdb_nsq
+        String blastn_path = "/usr/blastn_2.2.30+"
+        File funcotator_data_source
+        Boolean run_alignment_artifact_filter = false
+
+        # HPVDeepSeekTertiaryAnalysis inputs
```
kachulis, line 53: is this part of initial release?

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+        File high_risk_snps_hpv
+        File hpv16_sublineages
+
+        # HPVDeepSeekNormalization inputs
+        File regions
+        File gapdh_regions
+        File fp_regions
+        Float ml_plasma= 60.0
```
kachulis: remove default

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+        Float ng_cfdna = 20.0
```
kachulis: remove default

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+            top_hpv_num_duplex_reads = HPVDeepSeekGenotyping.top_hpv_num_duplex_reads,
+            top_hpv_duplex_coverage = HPVDeepSeekGenotyping.top_hpv_duplex_coverage,
+            secondary_hpv_types = HPVDeepSeekGenotyping.secondary_hpv_types,
+            low_risk_hpv_genotypes_detected = HPVDeepSeekGenotyping.low_risk_hpv_genotypes_detected,
+            cthpvdna_per_human_genome_equivalents = HPVDeepSeekNormalization.cthpvdna_per_human_genome_equivalents,
+            cthpvdna_count_per_ml_plasma = HPVDeepSeekNormalization.cthpvdna_count_per_ml_plasma,
+            cthpvdna_count_per_ng_cfdna = HPVDeepSeekNormalization.cthpvdna_count_per_ng_cfdna
+    }
+
+    output {
```
kachulis, line 138: There are a ton of outputs here.  Are they all necessary, or can some be removed?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        Int min_ssd_size_gb = 512
+        Boolean use_ssd = true
+    }
+
+    String strategy = if is_duplex then "paired" else "adjacency"
+    String output_type = if is_duplex then "duplex" else "simplex"
+
+    command <<<
+        gatk MergeBamAlignment \
+        --ATTRIBUTES_TO_RETAIN X0 \
```
kachulis: what is this `X0` attribute being retained?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        --OUTPUT ~{output_basename}.merged.bam \
+        --REFERENCE_SEQUENCE ~{reference} \
+        --SORT_ORDER queryname \
+        --ALIGNED_READS_ONLY true \
+        --MAX_INSERTIONS_OR_DELETIONS -1 \
+        --PRIMARY_ALIGNMENT_STRATEGY MostDistant \
+        --ALIGNER_PROPER_PAIR_FLAGS true \
+        --CLIP_OVERLAPPING_READS false
+
+        samtools view -f 2 -q 1 -bh ~{output_basename}.merged.bam > ~{output_basename}.merged.filtered.bam
```
kachulis: you can use `-o` instead of redirecting stdout.

HPVDeepSeek/HPVDeepSeekSomaticVariantCalling.wdl
```diff
@@ -0,0 +1,624 @@
...
+        Int disk_size_gb = 128
+        Int min_ssd_size_gb = 512
+        Boolean use_ssd = true
+    }
+
+    command <<<
+        gatk --java-options "-Xms8g -Xmx14g" \
+        FilterAlignmentArtifacts \
+        --input ~{bam} \
+        --output ~{output_basename}.vcf.gz
```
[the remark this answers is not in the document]
kockan: Fixed

HPVDeepSeek/HPVDeepSeekSomaticVariantCalling.wdl
```diff
@@ -0,0 +1,624 @@
...
+        }
+    }
+
+    File mapping_filter_input_vcf = select_first([FilterAlignmentArtifacts.filtered_vcf, FilterMutectCalls.filtered_vcf])
+    File mapping_filter_input_vcf_idx = select_first([FilterAlignmentArtifacts.filtered_vcf_idx, FilterMutectCalls.filtered_vcf_idx])
+
+    call RunMappingFilter {
+        input:
+            vcf = mapping_filter_input_vcf,
+            vcf_idx = mapping_filter_input_vcf,
```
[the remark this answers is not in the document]
kockan: Fixed

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        input:
+            bam = AlignReads.bam
+    }
+
+    if(downsample) {
+        call Downsample {
+            input:
+                bam = SortAndIndexBam.sorted_bam,
+                bai = SortAndIndexBam.sorted_bam_index,
+                downsample_probability = downsample_probability,
```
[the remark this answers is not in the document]
kockan: This is okay

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
...
+                if not region.startswith("HPV"):
+                    hg38_simplex_reads = hg38_simplex_reads + int(columns[1])
+                    hg38_duplex_reads = hg38_duplex_reads + int(columns[2])
+
+        num_non_xy_fp_regions = 0
+        for fp_region in fp_region_list:
+            if (not fp_region.startswith("chrX")) and (not fp_region.startswith("chrY")):
+                num_non_xy_fp_regions += 1
+
+        mean_simplex_depth_hg38_fp_only = mean_simplex_depth_hg38_fp_only / num_non_xy_fp_regions
```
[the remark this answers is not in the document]
kockan: Should never happen but fair suggestion and doesn't hurt. Fixed.

HPVDeepSeek/HPVDeepSeekTertiaryAnalysis.wdl
```diff
@@ -0,0 +1,244 @@
...
+        String genotype = "HPV16_Ref"
+
+        Int cpu = 2
+        Int memory_gb = 16
+        Int disk_size_gb = ceil((2.5 * size(bam, "GiB")) + 50)
+    }
+
+    command <<<
+        bcftools mpileup -Ou --fasta-ref ~{reference} --max-depth 800000 --regions ~{genotype} --annotate FORMAT/AD,INFO/AD,FORMAT/AD,FORMAT/ADF ~{bam} \
+        | bcftools call -Ou -mv -o ~{output_basename}.vcf.gz
```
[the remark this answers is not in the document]
kockan: This is fine, it's an internal output in the task anyway and is handled properly. Might refactor later as needed.

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
...
+            header = f.readline()
+            columns = f.readline().rstrip().split('\t')
+
+            num_hpv_simplex_reads_fs_geq_5 = int(columns[5])
+            num_hg38_simplex_reads_fp_only = int(columns[9])
+            mean_simplex_depth_hg38_fp_only = float(columns[11])
+            mean_simplex_depth_hpv_fs_geq_5 = float(columns[16])
+
+            cthpvdna_per_human_genome_equivalents = mean_simplex_depth_hpv_fs_geq_5 / mean_simplex_depth_hg38_fp_only
+            cthpvdna_per_ml_plasma = num_hpv_simplex_reads_fs_geq_5 / ~{ml_plasma}
```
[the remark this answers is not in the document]
kockan: Should never happen and perhaps I'd rather this crash the workflow since that would mean something is very wrong.

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+            f.write(max_elem[0])
+
+        with open("top_hpv_num_duplex_reads.txt", 'w') as f:
+            f.write(str(max_elem[1][0]))
+
+        with open("top_hpv_duplex_coverage.txt", 'w') as f:
+            f.write(str(max_elem[1][1]))
+
+        with open("is_hpv_positive.txt", 'w') as f:
+            if max_elem[1][0] >= 4 and max_elem[1][1] >= 5.0:
```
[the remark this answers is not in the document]
kockan: Resolved

kockan:
@kachulis Another round?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+# --REFERENCE_SEQUENCE reference.fasta \
+# --SORT_ORDER coordinate \
+# --ATTRIBUTES_TO_RETAIN X0 \
+# --ATTRIBUTES_TO_RETAIN RX \
+# --ADD_MATE_CIGAR true \
+# --MAX_INSERTIONS_OR_DELETIONS -1 \
+# --PRIMARY_ALIGNMENT_STRATEGY MostDistant \
+# --ALIGNER_PROPER_PAIR_FLAGS true \
+# --CLIP_OVERLAPPING_READS false
+task MergeConsensus {
```
kachulis: thoughts on refactoring so that there's a single MBA task called for both consensus and raw, with the filtering and umi grouping split out to a separate task?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        String output_basename
+        File bam
+
+        Int cpu = 2
+        Int memory_gb = 16
+        Int disk_size_gb = ceil((2 * size(bam, "GiB")) + 50)
+    }
+
+    command <<<
+        samtools coverage ~{bam} | awk '{if(NR==1){printf "%s\t%s\t%s\n",$1,$4,$6} else {printf "%s\t%s\t%s\n",$1,$4,$6}}' > ~{output_basename}.coverage.txt
```
kachulis: the `printf` statements appear to be the same here, is the `if` really needed?

And am I correct that this is just grabbing the 1st, 4th, and 6th columns from the samtools coverage output?  In which case you could just use `cut`?  But also, why not just keep the output in the format samtools produces?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        docker: "us-central1-docker.pkg.dev/broad-gp-hydrogen/hydrogen-dockers/kockan/hds@sha256:56f964695f08ddb74e3a29c63c3bc902334c1ddd735735cc98ba6d6a4212285c"
+    }
+}
+
+# HPV+ Classification
+# Sample is considered HPV+ if the following thresholds are met:
+# Duplex read count ≥ 4
+# Percentage of HPV genome covered by read alignment ≥ 5%
+# The HPV genotype with the highest number of duplexes is NOT a low-risk HPV genotype
+task DetermineHPVStatus {
```
kachulis, line 595: I would push to move away from the framing of "top_hpv_genotype" here.  Instead, just have this task output a single file.  Output file contains a table which is all the hpv strains with any duplex reads, number of duplex reads, % of HPV strain covered, and whether or not strain is called as "detected" (basically, is it a "high risk" strain.  So something like:

```
hpv_strain, n_duplex, frac_cov_duplex, detected
HPV16, 25, 0.7, true
HPV59, 372, 1.0, true
HPV6, 18, 0.4, false
```

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        set -e
+        python3 <<CODE
+
+        low_risk_hpv_genotype_list = []
+        with open("~{low_risk_hpv_genotypes}", 'r') as f:
+            low_risk_hpv_genotype_list = f.read().splitlines()
+
+        coverage_dict = {}
+        low_risk_hpv_dict = {}
+        with open("~{coverage}", 'r') as infile:
```
kachulis: I think this whole thing could just be reading into a pandas df and then one or two simple pandas operations.

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+        Int top_hpv_num_duplex_reads = read_int("top_hpv_num_duplex_reads.txt")
+        Float top_hpv_duplex_coverage = read_float("top_hpv_duplex_coverage.txt")
+        Boolean is_hpv_positive = read_boolean("is_hpv_positive.txt")
+        String secondary_hpv_types = read_string("secondary_hpv_types.txt")
+        String low_risk_hpv_genotypes_detected = read_string("low_risk_hpv_genotypes_detected.txt")
+    }
+}
+
+# Human SNP Genotyping
+task GenotypeSNPsHuman {
```
kachulis, line 642: is this needed for this version?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+            read_group_description = read_group_description,
+            output_basename = output_basename
+    }
+
+    call SortAndIndexBam {
+        input:
+            bam = AlignReads.bam
+    }
+
+    if(downsample) {
```
kachulis: I understand this was helpful to have during development for analysis purposes.  But I think this WDL should be trimmed down to just what will be done for HPVDeepSeek production.  

Can keep these types of analytical add-ons on a separate branch.

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+            bam = aligned_bam
+    }
+
+    call CollectInsertSizeMetrics as PreConsensusInsertSizeMetrics {
+        input:
+            bam = aligned_bam,
+            bai = aligned_bam_index
+    }
+
+    call CountOnTargetReads as PreConsensusCountOnTargetReads {
```
kachulis: why?  Don't you get this from CollectHSMetrics?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+            bam = SortAndIndexSimplexBam.sorted_bam
+    }
+
+    call CollectInsertSizeMetrics as PostConsensusInsertSizeMetrics {
+        input:
+            bam = SortAndIndexSimplexBam.sorted_bam,
+            bai = SortAndIndexSimplexBam.sorted_bam_index
+    }
+
+    call CountOnTargetReads as PostConsensusCountOnTargetReads {
```
kachulis: needed?

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -0,0 +1,1556 @@
...
+    call CollectAlignmentSummaryMetrics as PreConsensusAlignmentSummaryMetrics {
+        input:
+            bam = aligned_bam,
+            bai = aligned_bam_index,
+            reference = reference,
+            reference_fai = reference_fai,
+            reference_dict = reference_dict
+    }
+
+    call Flagstat as PreConsensusFlagstat {
```
kachulis: is this useful?

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+            read_group_id = read_group_id,
+            read_group_sample_name = read_group_sample_name,
+            read_group_library_name = read_group_library_name,
+            read_group_platform = read_group_platform,
+            read_group_platform_unit = read_group_platform_unit,
+            read_group_description = read_group_description,
+            read_structure = read_structure
+    }
+
+    call HPVDeepSeekSomaticVariantCalling.HPVDeepSeekSomaticVariantCalling {
```
kachulis: Not part of initial version, right?

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+            mapping_filter_python_script = mapping_filter_python_script,
+            blastdb_nhr = blastdb_nhr,
+            blastdb_nin = blastdb_nin,
+            blastdb_nsq = blastdb_nsq,
+            blastn_path = blastn_path,
+            funcotator_data_source = funcotator_data_source,
+            run_alignment_artifact_filter = run_alignment_artifact_filter
+    }
+
+    call HPVDeepSeekTertiaryAnalysis.HPVDeepSeekTertiaryAnalysis {
```
kachulis: not part of initial version, right?

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -0,0 +1,310 @@
...
+        File funcotator_data_source
+        Boolean run_alignment_artifact_filter = false
+
+        # HPVDeepSeekTertiaryAnalysis inputs
+        File high_risk_snps_hpv
+        File hpv16_sublineages
+
+        # HPVDeepSeekNormalization inputs
+        File regions
+        File gapdh_regions
```
kachulis: not needed anymore?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
...
+workflow HPVDeepSeekNormalization {
+    input {
+        String sample_id
+        String top_hpv_genotype
+        File simplex_bam
+        File duplex_bam
+        File simplex_bam_index
+        File duplex_bam_index
+        File regions
+        File gapdh_regions
```
kachulis: is this still necessary?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
+version 1.0
+
+task FilterConsensusReads {
```
kachulis: this is still in format for when we were deciding on thresholds.  can be massively simplified now that thresholds have been choosen.

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
...
+
+    command <<<
+        set -e
+        python3 <<CODE
+
+        with open("~{fs_stats_summary}", 'r') as f:
+            header = f.readline()
+            columns = f.readline().rstrip().split('\t')
+
+            num_hpv_simplex_reads_fs_geq_5 = int(columns[5])
```
kachulis: don't need just counts anymore, right?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,437 @@
...
+
+    runtime {
+        cpu: cpu
+        memory: "~{memory_gb} GiB"
+        disks: "local-disk ~{disk_size_gb} HDD"
+        docker: "us-central1-docker.pkg.dev/broad-gp-hydrogen/hydrogen-dockers/kockan/hds@sha256:56f964695f08ddb74e3a29c63c3bc902334c1ddd735735cc98ba6d6a4212285c"
+    }
+}
+
+workflow HPVDeepSeekNormalization {
```
kachulis, line 79: Normalization needs to be calculated for each detected hpv genotype separately

kachulis: I think this whole workflow should be simplified and updated to calculate normalized quantification result for each genotype separately

HPVDeepSeek/HPVDeepSeekSomaticVariantCalling.wdl
```diff
@@ -0,0 +1,624 @@
+version 1.0
```
kachulis, line 1: not yet relevant, can we move to a different branch?

HPVDeepSeek/HPVDeepSeekTertiaryAnalysis.wdl
```diff
@@ -0,0 +1,244 @@
+version 1.0
```
kachulis, line 1: not yet relevant, can we move to a different branch?

kachulis (changes requested):
Thanks Can, this is looking good!  A bunch of specific comments/requests, as well as a general comment:  Right now, this WDL still does a lot of things that were relevant during development for various analyses, but are not relevant to the actual production workflow.  Should be cleaned up to remove all that extra stuff (can be moved to a different branch so you have it in the future if you need it).  There are also a somatic variant calling and tertiary analysis workflow in here that aren't included in initial production, so should be move to a different branch for future dev, imo.

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Yes, but as unvalidated outputs as part of the Research product

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Yes, but as unvalidated outputs as part of the Research product

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: I believe `X0` is a bwa generated tag which is essentially the number of "best hits". So the ideal case would be 1, meaning the read only has a single mapping location with the highest alignment score.

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: For the Research product

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Got rid of some that we haven't really used so far

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Switched to this format now instead of a single top HPV genotype

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Downsampling removed

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Rewrote this function to generate the output in the format described in the previous comment. No pandas, but significantly simpler. We could still do it the pandas way if we want

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: We've never actually used/checked this metric. I believe there are better ways to collect it as you've mentioned. Removed

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Removed

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Not really, we have alternative ways to get the same/similar metrics. Removed

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Removed

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: This task has been completely replaced by another one, so no longer relevant

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: We don't use direct counts for normalization anymore, replaced by new version of subworkflow

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeek.wdl
kockan: Removed

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: I agree, this was part of the legacy pipeline from Faden Lab. The first line should be the header but there should be no need to treat it differently. Let's keep the output of samtools coverage as it and use the columns we want in the subsequent tasks

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Leftover from legacy pipeline. I agree the output flag is nicer. Done

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeek.wdl
```diff
@@ -38,7 +38,7 @@ workflow HPVDeepSeek {
         String read_structure = "3M2S+T"
 
         # HPVDeepSeekSomaticVariantCalling inputs
-        File target_intervals
+        File mutect_target_intervals
```
kachulis: what is the difference between `target_intervals` and `mutect_target_intervals`?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,243 @@
...
+            total_depth = 0
+            num_positions = 0
+
+            for pileupcolumn in infile_simplex.pileup(chromosome, start, end, stepper = "all", truncate = False, max_depth = 1000000, ignore_overlaps = True):
+                if chromosome.startswith("HPV"):
+                    for pileupread in pileupcolumn.pileups:
+                        if pileupread.alignment.get_tag("cD") >= 5:
+                            total_depth += 1
+                else:
+                    total_depth += pileupcolumn.nsegments
```
kachulis: shouldn't the hg38 depth also be calculated only from family size >=5 reads?  to avoided a bias between which reads are counted on hg38 vs which are counted on hpv?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,243 @@
...
+
+    runtime {
+        cpu: cpu
+        memory: "~{memory_gb} GiB"
+        disks: "local-disk ~{disk_size_gb} SSD"
+        docker: "us-central1-docker.pkg.dev/broad-gp-hydrogen/hydrogen-dockers/kockan/simple_pysam@sha256:a302f9efe0bf1d4f9998ee1e9dda406223454ccaea0b5619046742221c1d2a74"
+    }
+}
+
+task GetMedianOfHg38MeanDepthsSimplex {
```
kachulis: I don't think you need to split this out into a separate task.  do this w/ pandas and it's like 3 lines, isn't it?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,243 @@
...
+        Float ml_plasma
+        Float ng_cfdna
+    }
+
+    call CalculateMeanDepthsSimplex {
+        input:
+            sample_id = sample_id,
+            simplex_bam = simplex_bam,
+            simplex_bam_index = simplex_bam_index,
+            target_intervals = target_intervals
```
kachulis: why not just use fp_intervals here, and then you don't need to separately filter down to them later?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,243 @@
...
+        Int memory_gb = 16
+        Int disk_size_gb = 128
+    }
+
+    command <<<
+        set -e
+        python3 <<CODE
+
+        detected_hpv_genotypes = []
+        with open("~{hpv_status}", 'r') as f:
```
kachulis: any reason to parse this all yourself instead of just reading in w/ pandas?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -0,0 +1,243 @@
...
+                    detected_hpv_genotypes.append(columns[0])
+
+        median_hg38_val = 0.0
+        with open("~{median_hg38}", 'r') as f:
+            median_hg38_val = float(f.readline().strip())
+
+        outfile = open("~{sample_id}.normalized_hpv.tsv", 'w')
+
+        with open("~{hpv_target_mean_depths_fs_geq_5}", 'r') as f:
+            for line in f:
```
kachulis: again, why self parse instead of pandas?

kachulis (commented):
A few more comments about simplifying

HPVDeepSeek/HPVDeepSeek.wdl
kockan: A lot of the interval list and BED inputs needed refactoring here, this was actually a redundancy. Should be fixed in the next commit

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: We decided to apply the same family size filter to both hg38 and HPV. Resolved

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -609,38 +607,36 @@ task DetermineHPVStatus {
...
         python3 <<CODE
 
+        import pandas as pd
+
         low_risk_hpv_genotype_list = []
         with open("~{low_risk_hpv_genotypes}", 'r') as f:
             low_risk_hpv_genotype_list = f.read().splitlines()
 
         outfile = open("~{output_basename}.hpv_status.tsv", 'w')
         outfile.write("HPV_Genotype" + "\t" + "Num_Duplex_Reads" + "\t" + "%_Genomic_Coverage" + "\t" + "Is_Detected" + "\n")
```
kachulis: I'd kind of prefer to change "Is_Detected" to "Is_Reportable" or something.

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
```diff
@@ -609,38 +607,36 @@ task DetermineHPVStatus {
...
-                columns = line.split('\t')
+        df = pd.read_csv("~{coverage}", sep = '\t')
+        df = df.rename(columns={"#rname": "rname"})
 
-                chromosome = columns[0]
-                num_duplexes = int(columns[3])
-                genomic_coverage = float(columns[4])
+        df = df[(df["rname"].str.startswith("HPV")) & (df["numreads"] >= 2)]
+
+        for row in df.itertuples():
```
kachulis: I think this can be changed to 
```python
df['Is_reportable'] = ~df.rname.isin(low_risk_hpv_genotype_list)
```
and then rename the columns to what you want in the tsv, and write out the columns of interest to tsv w/ `df.to_csv`.  That way you don't need to explicitly write out the header above either.

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -1,187 +1,66 @@
...
             num_positions = 0
-
-            for pileupcolumn in infile_simplex.pileup(chromosome, start, end, stepper = "all", truncate = False, max_depth = 1000000, ignore_overlaps = True):
-                if chromosome.startswith("HPV"):
-                    for pileupread in pileupcolumn.pileups:
-                        if pileupread.alignment.get_tag("cD") >= 5:
-                            total_depth += 1
-                else:
-                    total_depth += pileupcolumn.nsegments
+            for pileupcolumn in infile_simplex.pileup(row.chromosome, row.start, row.end, stepper = "all", truncate = False, max_depth = 1000000, ignore_overlaps = True):
```
kachulis: I assume the max_depth = 1_000_000 should be safe in any reasonable use case here?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -1,187 +1,66 @@
...
+                for pileupread in pileupcolumn.pileups:
+                    if pileupread.alignment.get_tag("cD") >= 5:
+                        total_depth += 1
                 num_positions += 1
 
             mean_depth = 0.0
             if num_positions > 0:
```
kachulis: Do we want `num_positions == 0` to be handled and give mean_depth = 0?  Or should it result in an exception?

Put another way, if num_positions == 0 a sign of a problem, or is there a reasonable case where this could happen?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -1,187 +1,66 @@
...
-
-        median_hg38_val = 0.0
-        with open("~{median_hg38}", 'r') as f:
-            median_hg38_val = float(f.readline().strip())
+        mask = df["chromosome"].isin(df_detected_hpv_genotypes["HPV_Genotype"].tolist())
+        df.loc[mask, "r"] = df.loc[mask, "mean_depth"].apply(lambda x: x / hg38_median_depth)
+        df.loc[mask, "hpv_quantity"] = df.loc[mask, "r"].apply(lambda x: x * ~{ng_cfdna} / 0.0033 / ~{ul_plasma} / 1000.0)
 
         outfile = open("~{sample_id}.normalized_hpv.tsv", 'w')
+        outfile.write("HPV_Genotype" + "\t" + "HPV_Mean_Depth_Over_hg38_Median_Depth" + "\t" + "ng_cfDNA" + "\t" + "mL_Plasma" + "\t" + "HPV_Quantity" + "\n")
```
kachulis: `df.loc[mask].to_csv` instead?

HPVDeepSeek/HPVDeepSeekNormalization.wdl
```diff
@@ -1,187 +1,66 @@
...
-                if columns[0] not in detected_hpv_genotypes:
-                    continue
-
-                normalized_hpv_quantity = float(columns[3]) / median_hg38_val
-                normalized_hpv_quantity_ml_plasma = normalized_hpv_quantity / ~{ml_plasma}
-                normalized_hpv_quantity_ng_cfdna = normalized_hpv_quantity / ~{ng_cfdna}
-
-                outfile.write(columns[0] + "\t" + str(normalized_hpv_quantity) + "\t" + str(normalized_hpv_quantity_ml_plasma) + "\t" + str(normalized_hpv_quantity_ng_cfdna) + "\n")
-
+        infile_simplex.close()
```
kachulis: use a context manager instead?

kachulis (commented):
just a few more small things

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekGenotyping.wdl
kockan: Done

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: I don't believe this has been an issue in any samples we've processed so far. There might be rare cases where this might be an issue for samples with extremely high HPV read counts but definitely not for anything hg38

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: We can't really have a zero-length genomic interval, so `num_positions == 0` should never happen. For fingerprint sites the original interval length could be 1,  but that's still okay, and because we set `truncate = False` when using pysam pileup, in reality the length will very often be greater than 1 as long as it is covered by at least one simplex read.

So I suppose we can just get rid of this check since it might be better for the WDL to crash as a zero-length interval would mean something went really wrong.

kachulis (approved):
@kockan you can merge once comments are resolved.  Thanks!

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Refactored

HPVDeepSeek/HPVDeepSeekNormalization.wdl
kockan: Done

PR #243, opened 2026-05-05
Low GC Dropout metric for gcnv cohort wdl

kachulis:


gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
```diff
@@ -125,7 +131,86 @@ workflow CohortCNVCallingAndMergeForFabric {
...
+    runtime {
+        docker: "us.gcr.io/broad-dsde-methods/python-h5py@sha256:7f8d59658a06c585f005ee0cc33356ac029ce742cb561117f43a9345e3c05536"
+        preemptible: 3
+        cpu: 1
+        disks: "local-disk 50 HDD"
+        memory: mem_gb + " GB"
+    }
+
+    output {
+        File low_gc_drouput_tsv = "~{basename}_low_gc_dropout_qc.tsv"
```
michaelgatzen: `drouput` typo

gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
```diff
@@ -125,7 +131,86 @@ workflow CohortCNVCallingAndMergeForFabric {
         Array[Boolean] qc_passed = qc_passed_scatter
         Array[File] cnv_metrics = ExtractPoNFreqAnnotateFilterAndQC.cnv_metrics
         Array[File] cnv_event_report = GCNVVisualzation.cnv_event_report
+        Array[File] low_gc_dropout_metric = LowGCDropoutQC.low_gc_drouput_tsv
```
michaelgatzen: also here typo

gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
```diff
@@ -125,7 +131,86 @@ workflow CohortCNVCallingAndMergeForFabric {
...
+        import h5py
+        import pandas as pd
+        import numpy as np
+
+        def read_counts(file_path):
+            with h5py.File(file_path, 'r') as f:
+                # 1. Read contig names and create index
+                # .asstr() handles the common HDF5 byte-string issue in Python
+                contigs = pd.DataFrame({
+                    'CONTIG': f['intervals/indexed_contig_names'][:].astype(str)
```
michaelgatzen: So I actually spotted this and was confused because it seems to work in the tests, but then copilot also complained about it. Might be worth double checking?
> The comment says .asstr() handles HDF5 byte strings, but the code does not use .asstr(). With h5py, astype(str) on byte arrays often produces values like b'chr1' instead of chr1. If annotated_intervals has normal contig strings (chr1), this merge can fail or silently produce no matches.
>
> **Suggested fix**: use `f['intervals/indexed_contig_names'].asstr()[:]` or explicitly decode bytes.

gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
```diff
@@ -125,7 +131,86 @@ workflow CohortCNVCallingAndMergeForFabric {
...
+
+                # 7. Final cleanup (remove the internal index)
+                return counts_df.drop(columns=['contig_idx']), sample_name
+
+        counts_df, sample_name = read_counts("~{counts_hdf5}")
+        intervals = pd.read_csv("~{annotated_intervals}", sep='\t', comment="@")
+        counts_df = counts_df.merge(intervals)
+
+        low_gc_bin_counts = counts_df[(counts_df['GC_CONTENT'] > 0.25) & (counts_df['GC_CONTENT'] < 0.3)]['counts']
+        low_gc_dropout_frac = (low_gc_bin_counts < 0.25 * low_gc_bin_counts.mean()).mean()
```
michaelgatzen, line 193: Is there ever a case where `low_gc_bin_counts` could be empty? If so I think that would cause trouble with the `mean()`

michaelgatzen (approved):
So this is the first time I've used Copilot to help me with the review and I have to say, it did find the one thing I was unsure about, and a few more things. I would not have found the typo I think.

Anyway, I'm going to approve it because the tests seem to be passing. You can take a look at the comments and then decide what you want to change. If you do want another review just let me know.

gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
kachulis: oh interesting.  Yes it does seem to work w/ `.astype(str)`, but I'll switch it to `.asstr()` to be safe.

gCNV/cohort_cnv_calling_and_merge_for_fabric.wdl
kachulis: In theory yes,  `low_gc_bin_counts` could be empty, though it generally shouldn't happen in standard usage.  I think the result would just be that `low_gc_droupout_frac` would be `nan`, which I think is reasonable behavior.

